PR #3, opened 2015-08-30
Document everything, remove unused error codes

erickt:


erickt:
@oli-obk: by the way, this and [serde](https://github.com/serde-rs/serde) are now setup to use https://homu.io.

PR #24, opened 2015-12-12
Add helper function to parse stream of json values

little-arhat:
Many modern web APIs return their data as stream of (sometimes newline
delimited) JSON values. This commits add handy function `parse_stream`
that returns Iterator over parsed values.

This feature has some drawbacks or issues:

* I'm not sure, whether it's good idea to expose `JSONStream` type, but it seems you cannot currently return abstract `Iterator`
* Ideally, we should accept `io::Read`, but I couldn't get it to work, unfortunately: `expected 'Iter', found `std::io::Bytes<R>`
* The hardest one — naming, — not sure, how to name iterator and helper function. 

Commit contains tests (and they are passing!) 

WDYT?

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -6,7 +6,7 @@ use std::char;
 use std::i32;
 use std::io;
 use std::str;
-use core::marker::PhantomData;
+use std::marker::PhantomData;
 
 use serde::de;
 use serde::iter::LineColIterator;
@@ -830,7 +830,7 @@ pub fn from_str<T>(s: &str) -> Result<T>
     from_slice(s.as_bytes())
 }
 
-/// Iterator over JSON values
+/// Iterator over parsed JSON values
 pub struct JSONStream<T, Iter>
     where Iter: Iterator<Item=io::Result<u8>>,
           T: de::Deserialize
@@ -839,10 +839,13 @@ pub struct JSONStream<T, Iter>
     _marker: PhantomData<T>,
 }
 
-impl <T, Iter>JSONStream<T, Iter>
+/// Iterator over parsed JSON values
+impl <T, Iter> JSONStream<T, Iter>
     where Iter:Iterator<Item=io::Result<u8>>,
           T: de::Deserialize {
-    fn new(i: Iter) -> JSONStream<T, Iter> {
+    /// Returns Iterator of decoded JSON value from an iterator over
+    /// `Iterator<Item=io::Result<u8>>`.
+    pub fn new(i: Iter) -> JSONStream<T, Iter> {
         JSONStream {
             deser: Deserializer::new(i),
             _marker: PhantomData
@@ -850,32 +853,20 @@ impl <T, Iter>JSONStream<T, Iter>
     }
 }
 
-impl <T, Iter>Iterator for JSONStream<T, Iter>
+impl <T, Iter> Iterator for JSONStream<T, Iter>
     where Iter:Iterator<Item=io::Result<u8>>,
           T: de::Deserialize {
     type Item = Result<T>;
     fn next(&mut self) -> Option<Result<T>> {
-        match de::Deserialize::deserialize(&mut self.deser) {
-            Ok(v) => Some(Ok(v)),
-            Err(e) => {
-                match e {
-                    Error::SyntaxError(
-                        ErrorCode::EOFWhileParsingValue, _, _) =>
-                        match self.deser.end() {
-                            Ok(_) => None,
-                            Err(e) => Some(Err(e))
-                        },
-                    _ => Some(Err(e))
-                }
-            }
+        match self.deser.eof() {
+            Ok(true) => None,
+            Ok(false) => match de::Deserialize::deserialize(&mut self.deser) {
+                Ok(v) => Some(Ok(v)),
+                // EOF is handled beforehand, so report any error
+                Err(e) => Some(Err(e))
+            },
+            // Should not happen, seek .eof()
+            Err(e) => Some(Err(e))
         }
     }
-}
-
-/// Returns Iterator of decoded JSON value from an iterator over
-/// `Iterator<Item=io::Result<u8>>`.
-pub fn parse_stream<T, Iter>(i: Iter) -> JSONStream<T, Iter>
-    where Iter: Iterator<Item=io::Result<u8>>,
-          T: de::Deserialize {
-    JSONStream::new(i)
 }
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -858,15 +858,19 @@ impl <T, Iter> Iterator for JSONStream<T, Iter>
           T: de::Deserialize {
     type Item = Result<T>;
     fn next(&mut self) -> Option<Result<T>> {
-        match self.deser.eof() {
-            Ok(true) => None,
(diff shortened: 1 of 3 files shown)


little-arhat:
should I be worried about failing build on `stable`? I currently can't even build stable myself:

```
/Users/u/.multirust/toolchains/stable/cargo/registry/src/github.com-0a35038f75765ae4/quasi-0.3.10/src/lib.rs:11:43: 11:66 error: #[feature] may not be used on the stable release channel
/Users/u/.multirust/toolchains/stable/cargo/registry/src/github.com-0a35038f75765ae4/quasi-0.3.10/src/lib.rs:11 #![cfg_attr(not(feature = "with-syntex"), feature(rustc_private))]
```

little-arhat:
@erickt how does it look? should I fix it for stable? I can't compile crate for stable on my machine (maybe should do some clean), but error from travis ci is `error: use of unstable library feature 'core': the libcore library has not yet been scrutinized for stabilization in terms of structure and naming`. And as far as I know, one can't simply use feature flags on stable.

erickt:
@little-arhat: Sorry for the delay! I'll review it now.

json/src/de.rs
```diff
@@ -6,6 +6,7 @@ use std::char;
 use std::i32;
 use std::io;
 use std::str;
+use core::marker::PhantomData;
```
erickt: You should use `use std::marker::PhantomData` here.

json/src/de.rs
```diff
@@ -828,3 +829,53 @@ pub fn from_str<T>(s: &str) -> Result<T>
...
+          T: de::Deserialize
+{
+    deser: Deserializer<Iter>,
+    _marker: PhantomData<T>,
+}
+
+impl <T, Iter>JSONStream<T, Iter>
+    where Iter:Iterator<Item=io::Result<u8>>,
+          T: de::Deserialize {
+    fn new(i: Iter) -> JSONStream<T, Iter> {
```
erickt: This should be public, and also have a comment.

json/src/de.rs
```diff
@@ -828,3 +829,53 @@ pub fn from_str<T>(s: &str) -> Result<T>
...
+/// Iterator over JSON values
+pub struct JSONStream<T, Iter>
+    where Iter: Iterator<Item=io::Result<u8>>,
+          T: de::Deserialize
+{
+    deser: Deserializer<Iter>,
+    _marker: PhantomData<T>,
+}
+
+impl <T, Iter>JSONStream<T, Iter>
```
erickt: Please put a space here, as in `impl <T, Iter> JSONStream<T, Iter>`.

json/src/de.rs
```diff
@@ -828,3 +829,53 @@ pub fn from_str<T>(s: &str) -> Result<T>
...
+    where Iter:Iterator<Item=io::Result<u8>>,
+          T: de::Deserialize {
+    type Item = Result<T>;
+    fn next(&mut self) -> Option<Result<T>> {
+        match de::Deserialize::deserialize(&mut self.deser) {
+            Ok(v) => Some(Ok(v)),
+            Err(e) => {
+                match e {
+                    Error::SyntaxError(
+                        ErrorCode::EOFWhileParsingValue, _, _) =>
```
erickt: I don't think this is the right thing do to here. It'll silently ignore a truncated json body. A more correct option would be to use a peek iterator, and before you parse the next json record, you can check if there's actually any bytes available to be read.

json/src/lib.rs
```diff
@@ -125,6 +125,8 @@ pub use self::de::{
     from_reader,
     from_slice,
     from_str,
+    parse_stream,
+    JSONStream,
```
erickt: This should be put in the sorted location.

json_tests/tests/test_json.rs
```diff
@@ -14,6 +14,8 @@ use serde_json::{
     Value,
     from_str,
     from_value,
+    parse_stream,
+    JSONStream,
```
erickt: Same here.

json/src/de.rs
little-arhat: There is `pub struct Deserializer` and not public `impl<Iter> Deserializer<Iter>`, so I've followed this pattern. Should I change it to public?

json_tests/tests/test_json.rs
```diff
@@ -1393,3 +1395,19 @@ fn test_byte_buf_de() {
...
+                                                   .as_bytes()
+                                                   .iter()
+                                                   .map(|byte| Ok(*byte)));
+    assert_eq!(parsed.next().unwrap().ok().unwrap().lookup("x").unwrap(),
+               &Value::U64(40));
+    assert_eq!(parsed.next().unwrap().ok().unwrap().lookup("x").unwrap(),
+               &Value::U64(41));
+    assert_eq!(parsed.next().unwrap().ok().unwrap().lookup("x").unwrap(),
+               &Value::U64(42));
+    assert!(parsed.next().is_none());
```
erickt: Can you also add a test that check that handles if the last record is truncated, and make sure it returns an error?

json/src/de.rs
```diff
@@ -828,3 +829,53 @@ pub fn from_str<T>(s: &str) -> Result<T>
...
+        }
+    }
+}
+
+/// Returns Iterator of decoded JSON value from an iterator over
+/// `Iterator<Item=io::Result<u8>>`.
+pub fn parse_stream<T, Iter>(i: Iter) -> JSONStream<T, Iter>
+    where Iter: Iterator<Item=io::Result<u8>>,
+          T: de::Deserialize {
+    JSONStream::new(i)
```
erickt: You can remove this, and just rely on the constructor for `JSONStream`.

json/src/de.rs
erickt: Yes please just change the `new` method to be public. The `parse_stream` doesn't really add anything here, so it'd be easier to have users use the `new` constructor.

little-arhat:
@erickt thanks, fixed all remarks.

little-arhat:
@erickt hey, did you have time to look at updated code? Thanks!

little-arhat:
Fixed some corner cases found during using this.

Also, "All checks have passed" — @erickt what do you think? %)

json/src/de.rs
```diff
@@ -828,3 +829,48 @@ pub fn from_str<T>(s: &str) -> Result<T>
...
+          T: de::Deserialize {
+    type Item = Result<T>;
+    fn next(&mut self) -> Option<Result<T>> {
+        // skip whitespaces, if any
+        // this helps with trailing whitespaces, since whitespaces between
+        // values are handled for us.
+        let _:Result<()> = self.deser.parse_whitespace();
+        // Since Deserializer.eof() always return Ok(_), it's safe to
+        // call .ok() here
+        if self.deser.eof().ok().unwrap() {
```
nixpulvis: `Deserializer.eof` does not always return `Ok` there is a `try!` in there.

little-arhat:
@nixpulvis you're right, `Ok(try!` has confused me! :) Thanks, fixed.

little-arhat:
@erickt hi, any updates on that? %)

erickt:
@little-arhat: Sorry, I've been swamped with trying to get serde 0.7 out the door. That's *almost* across the line so I'm going to all these outstanding PRs again!

PR #25, opened 2015-12-31
Remove the Sized bound on to_value<T>()

tomjakubowski:
This allows to_value() to accept &str arguments directly.


tomjakubowski:
Travis failure seems unrelated (and only on Nightly).

erickt:
A very belated thanks!

PR #26, opened 2016-01-05
Report correct error code for missing field

### LINKED ISSUE #22: Bad error code propagated for missing field when deserializing a struct
When deserializing a structure from a JSON string that's missing a field, `serde_json` propagates an `ExpectedSomeValue` error code variant, not the `MissingField` error code variant.

The program below shows the behavior.

Expected output (roughly speaking):

    Got: "missing field x" at line 3 column 5
    OK, got 'missing field' x

Actual output:

    Got: "expected value" at line 3 column 5
    BAD, got 'expected some value'

Here's the program:

    extern crate serde;
    extern crate serde_json;

    #[derive(Debug)]
    struct Foo {
        x: i32,
        y: i32,
    }

        fn deserialize<D>(d: &mut D) -> Result<Self, D::Error> where D: serde::Deserializer {

            enum Field {
                X,
                Y,
            }

            impl serde::Deserialize for Field {
                fn deserialize<D>(d: &mut D) -> Result<Self, D::Error> where D: serde::Deserializer {

                    struct Visitor;

                    impl serde::de::Visitor for Visitor {
                        type Value = Field;

                        fn visit_str<E>(&mut self, value: &str) -> Result<Self::Value, E>
                            where E: serde::de::Error
                        {
                            match value {
                                "x" => Ok(Field::X),
                                "y" => Ok(Field::Y),
                                _ => Err(E::unknown_field(value)),
                            }
                        }
                    }

                    d.visit(Visitor)
                }
            }

            struct Visitor;

            impl serde::de::Visitor for Visitor {
                type Value = Foo;

                fn visit_map<V>(&mut self, mut visitor: V) -> Result<Foo, V::Error>
                    where V: serde::de::MapVisitor
                {
                    let mut x = None;
                    let mut y = None;

                    loop {
                        match try!(visitor.visit_key()) {
                            Some(Field::X) => {
                                x = Some(try!(visitor.visit_value()));
                            }
                            Some(Field::Y) => {
                                y = Some(try!(visitor.visit_value()));
                            }
                            None => { break; }
                        }
                    }

                    try!(visitor.end());

                    let x = match x {
                        Some(x) => x,
                        None => try!(visitor.missing_field("x")),
                    };

                    let y = match y {
                        Some(x) => x,
                        None => try!(visitor.missing_field("y")),
                    };

                    let v = Foo {
                        x: x,
                        y: y,
                    };

                    Ok(v)
                }
            }

            static FIELDS: &'static [&'static str] = &["x", "y"];
            d.visit_struct("Foo", FIELDS, Visitor)
        }
    }

    fn main() {

        let s = r#"{
            "y": 17
        }"#;

        let got = serde_json::from_str::<Foo>(&s).unwrap_err();

        println!("Got: {}", got);

        match got {
            serde_json::Error::SyntaxError(code, _line, _column) => {
                match code {
                    serde_json::ErrorCode::MissingField(field) => {
                        println!("OK, got 'missing field' {}", field);
                    }
                    serde_json::ErrorCode::ExpectedSomeValue => {
                        println!("BAD, got 'expected some value'");
                    }
                    _ => { println!("BAD, got some other error code variant"); }
                }
            }
            _ => { println!("BAD, Got unexpected error variant"); }
        }
    }

cmbrandenburg:
Fixes #22

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -745,11 +745,10 @@ impl<'a, Iter> de::MapVisitor for MapVisitor<'a, Iter>
         }
     }
 
-    fn missing_field<V>(&mut self, _field: &'static str) -> Result<V>
+    fn missing_field<V>(&mut self, field: &'static str) -> Result<V>
         where V: de::Deserialize,
     {
-        let mut de = de::value::ValueDeserializer::into_deserializer(());
-        Ok(try!(de::Deserialize::deserialize(&mut de)))
+        Err(Error::SyntaxError(ErrorCode::MissingField(field), 0, 0))
     }
 }
 


cmbrandenburg:
Fixes #29.

cmbrandenburg:
What's holding up this pull request? The build failures appear to have nothing to do with my changes. Is there some other problem? It would be nice to have some feedback.

JohnHeitmann:
Bystander comment: Thanks for this patch. I've run into this opaque error too.

Run a cargo test pass with rust nightly in the json_tests crate. You'll see some tests that are broken with this change. After a quick scan it looks like the behavior is better than before, and the tests just need updating.

I don't understand what the old code was trying to do, so I can't give good direct feedback.

cmbrandenburg:
@JohnHeitmann Thanks for the help. I learned two things:

1. The Cargo command for building `serde_json` with Rust Nightly is different than building with Rust Stable. (The Nightly command is: `cargo test --features with-syntex --no-default-features`.) Using the correct command, I can reproduce the test failures.

1. Looks like the `missing_field` method is supposed to provide a default value if one exists (e.g., `None` for an `Option` type), _otherwise_ report an error. Consequently, my pull request here is wrong. I must dig deeper to find the root of the problem where `missing_field` is reporting an incorrect error code.

PR #27, opened 2016-01-05
Make the constructor methods publicly visible for the PrettyFormatter.

jnicholls:
Unable to set the indent constant for a new PrettyFormatter instance, nor reuse PrettyFormatter in an external crate outside of the serde_json::ser::Serializer.

If you kept these private on purpose because you weren't sure of the API choice for setting the indent, let me know.


erickt:
Oops, good catch! This was accidental.

PR #11, opened 2015-10-05
Forward `InvalidType` and `InvalidLength` errors and remove the redundant `Error` suffix.

tomprogrammer:
This change is necessary if https://github.com/serde-rs/serde/pull/166 is merged. More information in the linked PR.

PR #32, opened 2016-02-02
feat(serde-up) Support serde 0.7

jwilm:
**NOT READY** this needs changes before merging; Cargo.toml spec is hard coded to a git rev

Mappings from visit_* to de/serialize_* should be self explanatory. [serde(deny_unknown_fields)] was added to a test struct since it was checking for denial. A couple of error types were added to support new
serde de::value::Error variants. Finally, there are a bunch of throwaway changes pinning the serde version to a 0.7 commit. These should be removed once 0.7 lands in favor of a crates.io semver spec.

### DIFF
--- a/json/src/error.rs
+++ b/json/src/error.rs
@@ -10,6 +10,7 @@ use std::result;
 use std::string::FromUtf8Error;
 
 use serde::de;
+use serde::ser;
 
 /// The errors that can arise while parsing a JSON stream.
 #[derive(Clone, PartialEq)]
@@ -67,6 +68,15 @@ pub enum ErrorCode {
 
     /// Unexpected end of hex excape.
     UnexpectedEndOfHexEscape,
+
+    /// Invalid length
+    Length(usize),
+
+    /// Incorrect type from value
+    Type(de::Type),
+
+    /// Catchall for syntax error messages
+    Syntax(String),
 }
 
 impl fmt::Debug for ErrorCode {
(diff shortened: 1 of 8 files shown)


jwilm:
Hmm, tests passed on my machine. Will look into the failure later.

jwilm:
The build is now passing. The serde dependency will still need to be updated to 0.7 once it is published.

jwilm:
Tests are passing with latest master of serde

json/src/error.rs
```diff
@@ -92,6 +102,9 @@ impl fmt::Debug for ErrorCode {
             ErrorCode::MissingField(ref field) => write!(f, "missing field \"{}\"", field),
             ErrorCode::TrailingCharacters => "trailing characters".fmt(f),
             ErrorCode::UnexpectedEndOfHexEscape => "unexpected end of hex escape".fmt(f),
+            ErrorCode::Length(ref len) => write!(f, "incorrect value length {}", len),
```
tomprogrammer: These changes are already contained in PR #11 and are likely to conflict if that PR is merged.

jwilm:
Rebased on latest json#master and tested with latest serde#master. There's one test failing right now that I don't have time to chase down; It looks like the `deny_unknown_fields` attribute isn't working as expected. [Link to failing line](https://github.com/jwilm/json/blob/0.7-support/json_tests/tests/test_json.rs#L999).

tomprogrammer:
I also stumbled across that issue. Look at this https://github.com/tomprogrammer/json/commit/5eb789dc4d8ad4db013b3f8f9881b6009a50d69c commit. You can see I introduced a new `ErrorCode::UnkownVariant`. You can use it like

```rust
 ("{\"unknown\":[]}", Error::Syntax(ErrorCode::UnknownVariant("unknown".to_string()), 1, 10)),
```

tomprogrammer:
The interaction between your and my PR are a bit unfortunate though. I don't know how to solve this best.

jwilm:
@erickt just mentioned

> On serde_json, I've also got 0.7 almost merged in. 

So my branch may be moot. Still, any commit adding support for 0.7 will have similar conflicts.

erickt:
This has actually landed since serde 0.7 was released!

PR #35, opened 2016-02-06
Fix negative zero

nixpulvis:
Currently there is a fixme that was never fixed, which causes `-0` to explode with an overflow.


nixpulvis:
I have no idea why the coverage tests regressed from this change.

erickt:
Thanks! Don't worry too much about coveralls, it's not super precise in tracking line coverage.

PR #37, opened 2016-02-13
Specialize unit enum variants to only serialize variant name

fuchsnj:
this simplifies serialization/deserialization for unit enum variants which only need to serialize the variant name
example:
```rust
enum Animal{
   Cat,
   Dog
}
let animal = Animal::Dog;
```
`animal` will now be serialized as `"Dog"` instead of `{"Dog":[]}`

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -17,6 +17,7 @@ pub struct Deserializer<Iter: Iterator<Item=io::Result<u8>>> {
     rdr: LineColIterator<Iter>,
     ch: Option<u8>,
     str_buf: Vec<u8>,
+    key_only_enum_variant: bool
 }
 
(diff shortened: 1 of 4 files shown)


nixpulvis:
:+1:

erickt:
I'm not sure if it'd be appropriate in the deserializer to decide to serialize enums like this. Instead, I mentioned in https://github.com/serde-rs/serde/issues/251#issuecomment-188382905 an idea on how to allow the serializee to decide on whether or not it should be serialized as a string. Can we shift the discussion over to that issue?

dtolnay:
I agree that this is better addressed by https://github.com/serde-rs/serde/issues/251, so we can close this PR.

PR #41, opened 2016-02-27
feat(value): implement RFC6901 JSON Pointer

### LINKED ISSUE #9: Consider using JSON pointers
JSON values provide a convenience to look up nested values called [`Value.lookup`](http://serde-rs.github.io/json/serde_json/value/enum.Value.html#method.lookup). The path is given separated by dots. It is impossible to escape dots in key names with the current method. I propose to use JSON pointers as described in [RFC6901](https://tools.ietf.org/html/rfc6901) as they provide a standard way to access nested values.

If you want to use JSON pointers I can also implement the change.

pyfisch:
Add a method `pointer(&str)` to `Value`. Deprecate `lookup(&str)`.
Bump version to 0.7.1.

Closes #9 

Notes:
* It should be possible to add a `pointer_mut(&str)` method to Value. This would allow to add and modify values. (Maybe even a delete method) I failed to add such a method because of borrow checker.
* Should [RFC6902 JSON Patch](https://tools.ietf.org/html/rfc6902) be implemented or is this something for a separate crate? (Patch is based on Pointer)

### DIFF
--- a/json/Cargo.toml
+++ b/json/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "serde_json"
-version = "0.7.0"
+version = "0.7.1"
 authors = ["Erick Tryzelaar <erick.tryzelaar@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "A JSON serialization file format"
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -98,6 +98,8 @@ impl Value {
         Some(target)
     }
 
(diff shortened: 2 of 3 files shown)


dtolnay:
@erickt or @pcwalton is one of you reviewing this?

erickt:
This looks good to me! Thanks!

erickt:
@homu: r+ 6fa1af0

pyfisch:
Looks like a clippy build failure but not a bug of my code.

oli-obk:
@homu retry

PR #49, opened 2016-03-19
Use indoc for JSON document literals in test

dtolnay:
[Indoc](https://github.com/dtolnay/indoc) takes a string literal and un-indents it so the leftmost non-space character is in the first column. Let me know whether you think this is valuable.

*Before:*
```rust
concat!(
    "[\n",
    "  false,\n",
    "  null,\n",
    "  [\n",
    "    \"foo\\nbar\",\n",
    "    3.5\n",
    "  ]\n",
    "]"
)
```

*After:*
```rust
indoc!(r#"
    [
      false,
      null,
      [
        "foo\nbar",
        3.5
      ]
    ]"#
)
```

### DIFF
--- a/json_tests/Cargo.toml
+++ b/json_tests/Cargo.toml
@@ -6,16 +6,18 @@ build = "build.rs"
 
 [features]
 default = ["serde_macros"]
-with-syntex = ["syntex", "serde_codegen"]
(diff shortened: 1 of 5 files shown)


erickt:
Neat! That does look a lot cleaner now!

PR #61, opened 2016-05-02
fix(value): Missing field error when deserializing from Value

### LINKED ISSUE #50: Missing field in a struct is reported as an InvalidType error code when deserializing from Value
Serde reports the `ErrorCode::InvalidType` variant when deserializing from a `Value` and a field is missing. It should report the `ErrorCode::MissingField` variant—same as when deserializing from a JSON-formatted string.

The program below shows the problem.

```rust
extern crate serde;
extern crate serde_json;

#[derive(Debug)]
struct Foo {
    field_1: i32,
}

    fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
        where D: serde::Deserializer
    {
        enum Field {
            Field1,
        }

        impl serde::Deserialize for Field {
            fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
                where D: serde::Deserializer
            {
                struct Visitor;

                impl serde::de::Visitor for Visitor {
                    type Value = Field;

                    fn visit_str<E>(&mut self, value: &str) -> Result<Self::Value, E>
                        where E: serde::de::Error
                    {
                        match value {
                            "field_1" => Ok(Field::Field1),
                            _ => Err(E::unknown_field(value)),
                        }
                    }
                }

                deserializer.deserialize(Visitor)
            }
        }

        struct Visitor;

        impl serde::de::Visitor for Visitor {
            type Value = Foo;

            fn visit_map<Vis>(&mut self, mut visitor: Vis) -> Result<Self::Value, Vis::Error>
                where Vis: serde::de::MapVisitor
            {
                let mut field_1 = None;

                loop {
                    match try!(visitor.visit_key()) {
                        Some(Field::Field1) => {
                            field_1 = Some(try!(visitor.visit_value()));
                        }
                        None => {
                            break;
                        }
                    }
                }

                try!(visitor.end());

                let field_1 = match field_1 {
                    Some(field_1) => field_1,
                    None => try!(visitor.missing_field("field_1")),
                };

                Ok(Foo { field_1: field_1 })
            }
        }

        static FIELDS: &'static [&'static str] = &["field_1"];
        deserializer.deserialize_struct("Foo", FIELDS, Visitor)
    }
}

fn main() {
    let json_value = serde_json::builder::ObjectBuilder::new().unwrap();
    let json_encoded = serde_json::to_string(&json_value).unwrap();

    match serde_json::from_str::<Foo>(&json_encoded) {
        Err(serde_json::Error::Syntax(serde_json::ErrorCode::MissingField("field_1"), 1, 2)) => (),
        x @ _ => {
            panic!("Got unexpected result (from_str): {:?}", x);
        }
    }

    match serde_json::from_value::<Foo>(json_value) {
        Err(serde_json::Error::Syntax(serde_json::ErrorCode::MissingField("field_1"), _, _)) => (),
        x @ _ => {
            panic!("Got unexpected result (from_value): {:?}", x);
        }
    }
}
```

Expected output: (success)

Actual output:
```
thread '<main>' panicked at 'Got unexpected result (from_value): Err(Syntax(invalid type: Unit, 0, 0))', src/main.rs:91
Process didn't exit successfully: `target/debug/serde_json_test` (exit code: 101)
```

arcnmx:
The logic for treating missing fields differed between the utf8 deserializer and the `Value` one, and missing fields would be incorrectly reported as an "invalid type" error instead. This makes the two deserializers consistent. I believe the "deserialize from unit" logic was just left over from before `deserialize_option` was a thing, and therefore isn't necessary anymore?

Fixes #50

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -938,29 +938,33 @@ impl<'a> de::MapVisitor for MapDeserializer<'a> {
         }
     }
 
-    fn missing_field<V>(&mut self, _field: &'static str) -> Result<V, Error>
+    fn missing_field<V>(&mut self, field: &'static str) -> Result<V, Error>
         where V: de::Deserialize,
     {
-        // See if the type can deserialize from a unit.
-        struct UnitDeserializer;
+        use std;
 
-        impl de::Deserializer for UnitDeserializer {
-            type Error = Error;
+        struct MissingFieldDeserializer(&'static str);
 
-            fn deserialize<V>(&mut self, mut visitor: V) -> Result<V::Value, Error>
+        impl de::Deserializer for MissingFieldDeserializer {
+            type Error = de::value::Error;
+
+            fn deserialize<V>(&mut self, _visitor: V) -> std::result::Result<V::Value, Self::Error>
                 where V: de::Visitor,
             {
-                visitor.visit_unit()
+                let &mut MissingFieldDeserializer(field) = self;
+                Err(de::value::Error::MissingField(field))
             }
 
-            fn deserialize_option<V>(&mut self, mut visitor: V) -> Result<V::Value, Error>
+            fn deserialize_option<V>(&mut self,
+                                     mut visitor: V) -> std::result::Result<V::Value, Self::Error>
                 where V: de::Visitor,
             {
                 visitor.visit_none()
             }
         }
 
-        Ok(try!(de::Deserialize::deserialize(&mut UnitDeserializer)))
+        let mut de = MissingFieldDeserializer(field);
+        Ok(try!(de::Deserialize::deserialize(&mut de)))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -941,14 +941,12 @@ impl<'a> de::MapVisitor for MapDeserializer<'a> {
     fn missing_field<V>(&mut self, field: &'static str) -> Result<V, Error>
         where V: de::Deserialize,
     {
-        use std;
-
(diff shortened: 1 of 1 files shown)


mkroman:
Is there any reason not to merge this?

The issues with `quasi` and Rust nightly have since been resolved, and the tests passed for Rust stable and beta.

json/src/value.rs
```diff
@@ -938,29 +938,33 @@ impl<'a> de::MapVisitor for MapDeserializer<'a> {
...
 
-        impl de::Deserializer for UnitDeserializer {
-            type Error = Error;
+        struct MissingFieldDeserializer(&'static str);
 
-            fn deserialize<V>(&mut self, mut visitor: V) -> Result<V::Value, Error>
+        impl de::Deserializer for MissingFieldDeserializer {
+            type Error = de::value::Error;
+
+            fn deserialize<V>(&mut self, _visitor: V) -> std::result::Result<V::Value, Self::Error>
```
oli-obk: why is this `use std` + `std::result::Result` necessary, it looks like it worked before? This isn't generated code that might not have access to `Result` directly

arcnmx: oh sorry, it was just copied verbatim from https://github.com/serde-rs/json/blob/master/json/src/de.rs#L745 - can fix!

oli-obk:
whoa, what happened to our coverage here?

changes lgtm, but I need to investigate that coverage

oli-obk:
ah, nothing to see here, all good: https://users.rust-lang.org/t/psa-code-coverage-reported-by-travis-cargo-coveralls-coverage-may-suddenly-drop-become-more-accurate/5702

@homu r+

Issue #62, opened 2016-05-03
Improve ergonomics: ToValue, impl FromStr, ToString


s-panferov:
Recently I've moved my projects to Serde and I found it amazing, but it has several inconveniences:

1. `FromStr` trait is not implemented for `Value`, so I can't use `str.parse()`
2. rustc-serialize has [`ToJson`](https://doc.rust-lang.org/rustc-serialize/rustc_serialize/json/trait.ToJson.html) trait which improves ergonomics a lot.
3. ToString is not implemented for `Value`.

I've made several pull requests:

1. #64 - create and impl ToJson trait
2. #63 - impl Display and FromStr traits

I believe that this will improve ergonomics. This will also improve compatibility with `rustc-serialize` and simplify migration.

dtolnay:
Your two PRs have merged and I filed #71 to track ToString.

PR #64, opened 2016-05-03
feat(value): create and impl ToJson trait

s-panferov:
Refs #62

What do you think? Tests fail on `nightly`, but this seems unrelated.


json/src/value.rs
```diff
@@ -1001,3 +1001,15 @@ pub fn from_value<T>(value: Value) -> Result<T, Error>
...
+pub trait ToJson {
+    /// Converts the value of `self` to an instance of JSON
+    fn to_json(&self) -> Value;
+}
+
+impl<T: ?Sized> ToJson for T where T: ser::Serialize {
+    fn to_json(&self) -> Value {
+        to_value(&self)
+    }
+}
```
erickt: It looks like there isn't a trailing newline at the end of the file. Could you add one?

erickt:
This seems reasonable to me. r=erickt with my comment addressed.

json/src/value.rs
s-panferov: Sure

oli-obk:
@erickt: @dtolnay and I don't have push access to the json repo

oli-obk:
testing whether I have homu access. The travis failure is due to clippy

@homu r+

PR #65, opened 2016-05-04
Correctly escape ASCII control characters in strings

raphlinus:
This patch escapes ASCII control characters in the range 0x00...0x1f, in accordance with the JSON spec.

Fixes #51

PR #66, opened 2016-05-05
WIP: Allow different string escaping strategies

PaulGrandperrin:
This is a WIP trying to solve the problems exposed in Issue #51, PR #58 and PR #65.

TL;DR;
When serializing to JSON, some people might want:
- Only escaping characters as mandated by ECMA-404, with maximum performance.
- All Unicode control characters to be escaped (C0, DEL and C1), just like in Python and Ruby, with slightly degraded performance.
- All non-ASCII characters to be escaped, and therefore generating ASCII compatible JSONs, juste like in Python with the option `ensure_ascii =True`, with even more degraded performance.

This PR tries to address all these needs.


oli-obk:
The travis failure is legit:

> src/lib.rs:119:1: 119:21 error: #[feature] may not be used on the stable release channel
src/lib.rs:119 #![feature(unicode)]

dtolnay:
Closing due to inactivity. I would still be interested in offering a choice of what characters to escape if anyone can find a nice way to do it.

PR #76, opened 2016-06-05
Serialize unit variants as name only

dtolnay:
Ref https://github.com/serde-rs/serde/issues/251. This is a rebase of #37 with more tests to ensure backward-compatibility with the current format.

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -18,7 +18,6 @@ pub struct Deserializer<Iter: Iterator<Item=io::Result<u8>>> {
     rdr: LineColIterator<Iter>,
     ch: Option<u8>,
     str_buf: Vec<u8>,
-    key_only_enum_variant: bool,
 }
 
 macro_rules! try_or_invalid {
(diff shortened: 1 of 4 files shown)


dtolnay:
@erickt I know you were previously against #37 but based on the discussion in https://github.com/serde-rs/serde/issues/251 I think this is the right solution.

I haven't seen a compelling argument that `{"A": []}` is the correct way to format an unit variant.

Serde_codegen is doing the right thing by calling `serialize_unit_variant`, and `serde_json` is doing the wrong thing by having chosen a surprising and unhelpful format for unit variants.

dtolnay:
Rebased to fix the merge conflict with #80.

dtolnay:
Ping @erickt

dtolnay:
Erick's feedback on IRC was:

> I wonder if some of this logic should be moved into serde proper to avoid these runtime checks

dtolnay:
@erickt I removed the runtime checks by using two different VariantVisitor implementations.

erickt:
@homu: r+ f48e8d6

@dtolnay: looks great!

Issue #57, opened 2016-04-25
Unify the number variants in Value


dtolnay:
I am not convinced by the justification in https://github.com/serde-rs/json/issues/48#issuecomment-191194370. Instead of three `I64`, `U64`, `F64` variants I would much rather use one `Number` variant backed by an arbitrary-precision number. We can keep `as_i64`, `as_u64`, `as_f64` for convenience.

PR #79, opened 2016-06-11
Added Decimal support with the decimal crate, standarizes on a single Number value.

nubis:
Hi, this is my proposal for adding decimal support. Some important things to notice:

- This depends on https://github.com/serde-rs/serde/pull/375 getting merged first.
- It merges all numeric Values into a single Value::Number, that wraps a d128.
- Most of all the actual number parsing is delegated to the decimal crate. (decimal places, exponents, etc)
- Numbers are also serialized using d128 format, you'll notice some specs were changed to reflect that.
- It proposes a bump to 0.8.0 because it adds the dependency and changes the serialization format of long numbers.

### DIFF
--- a/json/Cargo.toml
+++ b/json/Cargo.toml
@@ -13,6 +13,7 @@ keywords = ["json", "serde", "serialization"]
 nightly-testing = ["clippy"]
 
 [dependencies]
-serde = "^0.7.0"
+serde = { path = "../../serde/serde" }
 num = { version = "~0.1.27", default-features = false }
 clippy = { version = "^0.*", optional = true }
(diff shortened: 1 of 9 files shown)


dtolnay:
Thanks for both PRs.

:+1: for unifying the number variants into `Value::Number` (#57).

:-1: for using d128: https://github.com/serde-rs/serde/pull/375#issuecomment-225375807. It would be a risky move for us to tie the public API of such a mainstream project to a crate that has not been vetted at all by the community.

dtolnay:
Closing in favor of https://github.com/serde-rs/rfcs/pull/1.

Issue #54, opened 2016-04-21
Add feature flag to preserve order of map entries


dtolnay:
Similar to https://github.com/chyh1990/yaml-rust/pull/16 - using [linked-hash-map](https://github.com/contain-rs/linked-hash-map) instead of BTreeMap in serde_json::Value. This would allow preserving the order of keys in the JSON form when deserializing to a Value, and controlling the order of keys in the JSON form when serializing from a Value.

dtolnay:
This was fixed in #80. It is in v0.7.2.

PR #80, opened 2016-06-15
add preserve_order feature

laktak:
I've added a PR for #54 

linked-hash-map needs to be released (probably version = "0.0.11") before this can be merged.

I had to use `linked_hash_map::Iter<'a, String, Value>` as a replacement for `btree_map::IntoIter<String, Value>`. This in turn required me to clone keys/values in `deserialize_enum` and `visit_key`. I'm not sure this is the best solution.

Also since I'm new to Rust and this is my very first PR please check my code thoroughly :)

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -53,18 +53,22 @@ use error::Error;
 /// Represents a key/value type.
 #[cfg(not(feature = "preserve_order"))]
 pub type Map<K, V> = BTreeMap<K, V>;
-
-/// Represents the IntoIter type.
-#[cfg(not(feature = "preserve_order"))]
-pub type MapIntoIter<K, V> = btree_map::IntoIter<K, V>;
-
 /// Represents a key/value type.
 #[cfg(feature = "preserve_order")]
 pub type Map<K, V> = linked_hash_map::LinkedHashMap<K, V>;
 
 /// Represents the IntoIter type.
+#[cfg(not(feature = "preserve_order"))]
+pub type MapIntoIter<K, V> = btree_map::IntoIter<K, V>;
+/// Represents the IntoIter type.
 #[cfg(feature = "preserve_order")]
 pub type MapIntoIter<K, V> = linked_hash_map::IntoIter<K, V>;
+
+#[cfg(not(feature = "preserve_order"))]
+type Visitor<K, T> = de::impls::BTreeMapVisitor<K, T>;
+#[cfg(feature = "preserve_order")]
+type Visitor<K, T> = linked_hash_map::serde::LinkedHashMapVisitor<K, T>;
+
 
 /// Represents a JSON value
 #[derive(Clone, PartialEq)]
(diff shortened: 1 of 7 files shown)


dtolnay:
I would strongly prefer to use `IntoIter` with linked-hash-map. I have a pull request at https://github.com/contain-rs/linked-hash-map/pull/54 that makes the backend changes required for `IntoIter`, and a proto-pull request at https://github.com/dtolnay/linked-hash-map/pull/1 that adds `IntoIter`. Let's find a way to get those into linked-hash-map before tackling #54.

laktak:
OK. If you like I can update this when it's ready.

dtolnay:
https://github.com/contain-rs/linked-hash-map/pull/54 has finally been merged, now waiting on https://github.com/contain-rs/linked-hash-map/pull/57.

dtolnay:
That one merged too, waiting on a release https://github.com/contain-rs/linked-hash-map/issues/56.

json/src/value.rs
```diff
@@ -824,8 +856,8 @@ impl de::Deserializer for Deserializer {
             Some(_) => Err(de::Error::invalid_type(de::Type::Map)),
             None => visitor.visit(VariantDeserializer {
                 de: self,
-                val: Some(value),
-                variant: Some(Value::String(variant)),
+                val: Some(value.clone()),
```
dtolnay: These clones are no longer required.

json/src/value.rs
```diff
@@ -971,8 +1003,8 @@ impl<'a> de::MapVisitor for MapDeserializer<'a> {
         match self.iter.next() {
             Some((key, value)) => {
                 self.len -= 1;
-                self.value = Some(value);
-                self.de.value = Some(Value::String(key));
+                self.value = Some(value.clone());
```
dtolnay: These clones are no longer required.

json/src/value.rs
```diff
@@ -571,11 +601,13 @@ impl ser::Serializer for Serializer {
     }
 
     #[inline]
+
+    #[inline]
```
dtolnay: Please remove the extra `#[inline]`.

json/src/value.rs
```diff
@@ -426,13 +446,23 @@ impl de::Deserialize for Value {
...
             }
 
+            #[cfg(not(feature = "preserve_order"))]
             #[inline]
             fn visit_map<V>(&mut self, visitor: V) -> Result<Value, V::Error>
                 where V: de::MapVisitor,
             {
                 let values = try!(de::impls::BTreeMapVisitor::new().visit_map(visitor));
                 Ok(Value::Object(values))
             }
```
dtolnay: I think this would be clearer as a single method containing:

```rust
#[cfg(not(feature = "preserve_order"))]
type Visitor<K, T> = de::impls::BTreeMapVisitor<K, T>;
#[cfg(feature = "preserve_order")]
type Visitor<K, T> = linked_hash_map::serde::LinkedHashMapVisitor<K, T>;

let values = try!(Visitor::new().visit_map(visitor));
Ok(Value::Object(values))
```

That makes it clear they are doing exactly the same thing just with different visitor types.

dtolnay:
I released [v0.7.2](https://github.com/serde-rs/json/releases/tag/v0.7.2) with this feature.

PR #83, opened 2016-06-25
Trim down the Travis build

dtolnay:
This takes us from 28 minutes to 5 minutes. The biggest savings is not compiling all of Clippy + Syntex in release mode for the benchmark.....

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ rust:
 - beta
 - nightly
 
+matrix:
+  include:
+    - rust: nightly
+      env: BENCH=1
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


dtolnay:
Related: https://github.com/rust-lang/cargo/issues/2234 Cargo builds build-dependencies (syntex) in release mode when building benchmarks.

.travis.yml
```diff
@@ -20,27 +25,20 @@ before_script:
...
-  popd
 - |
-    (cd json && travis-cargo build) &&
-    (cd json && travis-cargo test) &&
-    (cd json_tests && travis-cargo test -- --features with-syntex --no-default-features) &&
-    (cd json_tests && travis-cargo --only nightly test -- features nightly-testing) &&
-    (cd json_tests && travis-cargo --only nightly bench -- features nightly-testing) &&
-    (cd json_tests && travis-cargo --only nightly bench -- --features "nightly-testing with-syntex" --no-default-features) &&
-    (cd json && travis-cargo --only stable doc)
+    if [ -z $BENCH ]; then
```
oli-obk: I don't epeak bash... What does the -z do, and why is the bench flag set to 1 on nightly?

oli-obk:
Is there any point in running benchmarks on travis?

.travis.yml
dtolnay: `-z` means empty i.e. `BENCH.len() == 0`. We run four builds: the three release channels stable beta and nightly, and a fourth one that is nightly with an environment variable BENCH=1. The bench build was the longest so I split it out into it's own build to increase parallelism. The stable/beta/nightly builds all run the first part of this `if` statement. The BENCH=1 build runs the second part which is just `travis-cargo bench`.

.travis.yml
```diff
@@ -7,6 +7,11 @@ rust:
 - beta
 - nightly
```
dtolnay: This is where the first three builds are declared.

.travis.yml
```diff
@@ -7,6 +7,11 @@ rust:
...
 
+matrix:
+  include:
```
dtolnay: This is where the fourth build is declared.

dtolnay:
> Is there any point in running benchmarks on travis?

Basically to make sure it compiles. Also I guess to have a record of performance to look back on when trying to identify what commit introduced a performance regression - it is faster than running `cargo bench` locally. The way I have it, it adds no time to the Travis build because it runs in parallel with the other three builds.

oli-obk:
Iirc cargo test also compiles the benches. But having the travis logs for future reference is a good idea. Thanks for the detailed explanation

@homu r+

PR #85, opened 2016-06-25
Correctly escape ASCII control characters in strings

npgm:
This fixes fix #65 by raphlinus.

All I did was merge the test file correctly. 

In regard to similar requests #58 and #66, is there a preferred solution?


dtolnay:
Thanks @raphlinus and @npgm. I released 0.7.3 with this feature.

PR #86, opened 2016-06-25
Speed up integer and float printing

### LINKED ISSUE #84: serialize_u64 is slow
It currently looks like this:

```rust
fn serialize_u64(&mut self, value: u64) -> Result<()> {
    write!(&mut self.writer, "{}", value).map_err(From::from)
}
```

I tested the following variation and it made an astonishing 20% difference in the [bench_log](https://github.com/serde-rs/json/blob/57739801bf53793062c84c5f144d07ce2f126a5d/json_tests/benches/bench_log.rs) benchmarks (294 MB/s vs 244 MB/s).

```rust
if value == 0 {
    self.writer.write_all(b"0").map_err(From::from)
} else {
    let mut scratch = [0u8; 20];
    let mut len = 0;
    while value > 0 {
        scratch[19 - len] = (value % 10) as u8 + b'0';
        value /= 10;
        len += 1;
    }
    self.writer.write_all(&scratch[20-len..20]).map_err(From::from)
}
```

That is enough to motivate looking for a better solution, or contributing a fix to Rust for whatever is making `write!` slow.

dtolnay:
Fixes #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ fn main() {
     map.insert("y".to_string(), 2.0);
 
     let s = serde_json::to_string(&map).unwrap();
-    assert_eq!(s, "{\"x\":1,\"y\":2}");
+    assert_eq!(s, "{\"x\":1.0,\"y\":2.0}");
 
     let deserialized_map: Map<String, f64> = serde_json::from_str(&s).unwrap();
     assert_eq!(map, deserialized_map);
@@ -78,7 +78,7 @@ fn main() {
     let point = Point { x: 1.0, y: 2.0 };
 
     let s = serde_json::to_string(&point).unwrap();
-    assert_eq!(s, "{\"x\":1,\"y\":2}");
+    assert_eq!(s, "{\"x\":1.0,\"y\":2.0}");
 
     let deserialized_point: Point = serde_json::from_str(&s).unwrap();
     assert_eq!(point, deserialized_point);
--- a/json/Cargo.toml
+++ b/json/Cargo.toml
@@ -19,3 +19,4 @@ num-traits = "~0.1.32"
 clippy = { version = "^0.*", optional = true }
 linked-hash-map = { version = "0.0.11", optional = true }
 itoa = "0.1"
+dtoa = "0.1"
--- a/json/Cargo.toml
+++ b/json/Cargo.toml
@@ -18,3 +18,4 @@ serde = "^0.7.0"
 num-traits = "~0.1.32"
 clippy = { version = "^0.*", optional = true }
 linked-hash-map = { version = "0.0.11", optional = true }
+itoa = "0.1"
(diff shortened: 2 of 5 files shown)


oli-obk:
So if I understand your [implementation](https://github.com/dtolnay/itoa/blob/master/src/lib.rs) correctly, the only difference is that the original formatting implementation wrote to a temp buffer and then wrote to the target with an optional padding, while your implementation skips the padding and writes directly. Does llvm optimize the buffer away or what's the reason `itoa` is as fast as the direct impl you wrote in the issue?

r=me if there's no reason to rewrite `itoa` to directly write to the target.

oli-obk:
I also wonder if it were possible to fix the original impl in Rust to be as fast as your version by adding a conditional check for whether going through the padding function is necessary.

dtolnay:
> Does llvm optimize the buffer away or could it be optimized even more by writing directly?

It cannot write directly because it writes [backwards, least significant digit first](https://github.com/dtolnay/itoa/blob/01cb34a686a66269490f8b95acf16d3504a35ad6/src/lib.rs#L56).

> So if I understand your implementation correctly, the only difference is that the original formatting implementation wrote to a temp buffer and then wrote to the target with an optional padding, while your implementation skips the padding and writes directly.

They both write to a (stack-allocated) temp buffer. The only difference is that theirs [does all this stuff](https://github.com/rust-lang/rust/blob/b8214dc6c6fc20d0a660fb5700dca9ebf51ebe89/src/libcore/fmt/mod.rs#L906-L966) and mine doesn't.

> I also wonder if it were possible to fix the original impl in Rust to be as fast as your version by adding a conditional check for whether going through the padding function is necessary.

I filed https://github.com/dtolnay/itoa/issues/1 to follow up.

dtolnay:
I added a similar library for printing floats. Note that the new library adds ".0" onto floats that are whole numbers, so we get "3.0" instead of "3". Let me know if you find this disagreeable, although RapidJSON and Jackson both write "3.0".

dtolnay:
Also the new library writes `1.7976931348623157e308` instead of 179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 which seems like a good thing.

oli-obk:
> Note that the new library adds ".0" onto floats that are whole numbers, so we get "3.0" instead of "3". Let me know if you find this disagreeable, although RapidJSON and Jackson both write "3.0".

I have no particular opinion on that. We might be breaking people's unit tests, but it's perfectly valid json to add arbitrary zeros.

> which seems like a good thing.

yes, definitely

lgtm, but I think we should hold off with such a user-visible change and do it together with 0.8

PR #88, opened 2016-06-27
Speed up float printing

dtolnay:
The floating point parts of #86 which need to wait until 0.8.


dtolnay:
I pushed this to a v0.8.0-rc branch.

PR #89, opened 2016-06-28
Speed up string printing

dtolnay:
I am working on porting [this benchmark](https://github.com/miloyip/nativejson-benchmark#stringify-time) to Rust. This change makes about 20-30% difference in citm_catalog.json and twitter.json which are heavy on strings.

Side note: together with #87 and #88 this brings us within 10% of RapidJSON which is the fastest C/C++ implementation (as measured by that benchmark).

### DIFF
--- a/json/src/ser.rs
+++ b/json/src/ser.rs
@@ -488,7 +488,8 @@ impl<'a> Formatter for PrettyFormatter<'a> {
     }
 }
 
-/// Serializes and escapes a `&[u8]` into a JSON string.
+/// DEPRECATED. Will be removed in 0.8.0.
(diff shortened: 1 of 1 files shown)


oli-obk:
awesome!

Implementation lgtm. Have you played around with unchecked indexing and/or a `u8` -> `[u8; 2]` mapping instead of mapping the nibbles? It's not great for the caches, but they should be big enough to hold a 512 byte array.

@homu r+

dtolnay:
I have not, but at this point I would like to turn my attention to parsing. There is a lot more to improve there.

PR #97, opened 2016-07-05
Specialize parser for Iter=&[u8]

dtolnay:
This improves parser performance by 30-60% (that's right, more than 2x faster for many inputs). Performance when reading from other types of iterators is not affected.

The tests are going to require a Serde release containing https://github.com/serde-rs/serde/pull/422 and https://github.com/serde-rs/serde/pull/423.

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -9,14 +9,82 @@ use std::str;
 use std::marker::PhantomData;
 
 use serde::de;
-use serde::iter::LineColIterator;
 
 use super::error::{Error, ErrorCode, Result};
 
+use read::{self, Read};
+
+//////////////////////////////////////////////////////////////////////////////
+
 /// A structure that deserializes JSON into Rust values.
-pub struct Deserializer<Iter: Iterator<Item=io::Result<u8>>> {
-    rdr: LineColIterator<Iter>,
-    ch: Option<u8>,
+pub struct Deserializer<Iter>(DeserializerImpl<read::IteratorRead<Iter>>)
+    where Iter: Iterator<Item=io::Result<u8>>;
+
+impl<Iter> Deserializer<Iter>
+    where Iter: Iterator<Item=io::Result<u8>>,
+{
+    /// Creates the JSON parser from an `std::iter::Iterator`.
+    #[inline]
+    pub fn new(rdr: Iter) -> Self {
+        Deserializer(DeserializerImpl::new(read::IteratorRead::new(rdr)))
+    }
+
+    /// The `Deserializer::end` method should be called after a value has been fully deserialized.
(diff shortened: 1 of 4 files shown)


dtolnay:
There are some contortions to keep the new `Read` trait out of the public API and to not affect the type parameters on the `Deserializer` struct. This lets us replace the manual specialization with actual specialization once that feature is stable.

json_tests/tests/test_json.rs
```diff
@@ -675,26 +679,37 @@ fn test_parse_unusual_ok<T>(errors: Vec<(&str, T)>)
...
     }
 }
 
 // FIXME (#5527): these could be merged once UFCS is finished.
 fn test_parse_err<T>(errors: Vec<(&'static str, Error)>)
     where T: Debug + PartialEq + de::Deserialize,
 {
-    for (s, err) in errors {
-        match (err, from_str::<T>(s).unwrap_err()) {
+    for &(s, ref err) in &errors {
```
dtolnay: This is because `Error` is not `Copy` or `Clone` so I need a reference in order to test both the error from `from_str` and the one from `from_iter`.

json/src/de.rs
```diff
@@ -885,7 +927,7 @@ pub struct StreamDeserializer<T, Iter>
     where Iter: Iterator<Item=io::Result<u8>>,
           T: de::Deserialize
 {
-    deser: Deserializer<Iter>,
+    deser: DeserializerImpl<read::IteratorRead<Iter>>,
```
dtolnay: `StreamDeserializer` is not specialized. Probably if you are using `StreamDeserializer` you have a stream of things and not a `&str` anyway.

json/src/read.rs
```diff
@@ -0,0 +1,182 @@
...
+            self.index += 1;
+            Some(ch)
+        } else {
+            None
+        })
+    }
+
+    #[inline]
+    fn peek(&mut self) -> io::Result<Option<u8>> {
+        Ok(if self.index < self.slice.len() {
```
oli-obk: even more so here.

oli-obk:
Those improvements are impressive. Code looks good to me.

json/src/read.rs
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+        pos
+    }
+}
+
+impl<'a> Read for SliceRead<'a> {
+    #[inline]
+    fn next(&mut self) -> io::Result<Option<u8>> {
+        Ok(if self.index < self.slice.len() {
```
dtolnay: Well it is the logic of `self.slice.get(self.index).map(|ch| *ch)`. I tried that and it was consistently about 10% slower. Let me know if it is any different on your computer.

I will add a comment.

Issue #90, opened 2016-06-28
Unescaped tab, line feed, carriage return should not be accepted in strings


dtolnay:
From [JSON standard](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf):

> Insignificant whitespace is allowed before or after any token. The whitespace characters are: character tabulation (U+0009), line feed (U+000A), carriage return (U+000D), and space (U+0020). Whitespace is not allowed within any token, except that space is allowed in strings.

dtolnay:
These are the only checks from [JSON_checker](http://json.org/JSON_checker/) that we fail. Rustc-serialize correctly rejects unescaped whitespace in strings.

oli-obk:
Is it a problem if our parser/deserializer is more lenient than the standard, as long as our serializer produces correct json?

StefanoD:
@oli-obk So, you want to _guess_ what the sendet wanted to send you? Can be dangerous...

dtolnay:
I think we should aim to accept valid JSON and reject invalid JSON. I would make one exception which is I think it is okay for us to accept types other than list and map at the root level.

oli-obk:
but if accepting valid json requires additional code and conditions, it's additional code we need to maintain and test + it slows down the regular path. If the correct way is faster/easier (like with forbidding trailing commas), then it's fine with me.

maciejhirsz:
Kind of related, I've been looking at control characters:

> The control characters U+0000–U+001F and U+007F come from ASCII

`0x7F` is not marked as `U` in the LUT.

maciejhirsz:
I've been roaming around, since I'm looking for more universal testing suite for myself.

@dtolnay 
> I would make one exception which is I think it is okay for us to accept types other than list and map at the root level.

That's not an exception, both ECMA 404 and RFC 7159 say that JSON text has to conform to the grammar of a JSON value.

@oli-obk
> it's additional code we need to maintain and test + it slows down the regular path.

I've done this with a LUT and it didn't slow down the regular path at all. The logic is pretty trivial, there isn't much to maintain or test.

dtolnay:
This was fixed in #98/#100.

PR #98, opened 2016-07-05
Don't allow control characters in strings

maciejhirsz:
#90

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -29,6 +29,36 @@ macro_rules! try_or_invalid {
     }
 }
 
+const QU: u8 = 1; // quote
+const BS: u8 = 2; // backslash
+const CT: u8 = 4; // control character
+
+// Look up table of valid codepoints permitted in strings. The only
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! I'll take a look after work unless @oli-obk gets to it first.

oli-obk:
Logic wise this is good. Since the array is only used in a boolean manner, we might consider doing a bitmask check on two `u64`s instead of indexing an array, but let's leave that to benchmarking.

It needs a few regression tests for some of the situations it's catching now.

dtolnay:
According to [RFC7159](https://tools.ietf.org/html/rfc7159):

> All Unicode characters may be placed within the quotation marks, except for the characters that must be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

Based on this language, I would consider it incorrect to reject `\x7F` in a string.

maciejhirsz:
Will look into it after work, thanks!

dtolnay:
I folded this into #100.

PR #100, opened 2016-07-06
Specialize parse_string for Iter=&[u8]

dtolnay:


oli-obk:
so this makes #98 obsolete, right?

lgtm, waiting for travis

SylvKT:
How do I disable this? The manifest of some Minecraft mods that I'm parsing have control characters, and now my app crashes when it enounters one of those mods.

PR #102, opened 2016-07-07
Fix clippy lints

dtolnay:
Clippy was not running because the Travis build used the "clippy" feature and clippy was cfg'd with the "nightly-testing" feature.

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -57,7 +57,7 @@ pub type Map<K, V> = BTreeMap<K, V>;
 #[cfg(feature = "preserve_order")]
 pub type Map<K, V> = LinkedHashMap<K, V>;
 
-/// Represents the IntoIter type.
+/// Represents the `IntoIter` type.
 #[cfg(not(feature = "preserve_order"))]
 pub type MapIntoIter<K, V> = btree_map::IntoIter<K, V>;
 /// Represents the IntoIter type.
@@ -101,8 +101,8 @@ impl Value {
     /// If the `Value` is an Object, returns the value associated with the provided key.
     /// Otherwise, returns None.
     pub fn find<'a>(&'a self, key: &str) -> Option<&'a Value>{
-        match self {
-            &Value::Object(ref map) => map.get(key),
+        match *self {
+            Value::Object(ref map) => map.get(key),
             _ => None
         }
     }
@@ -158,7 +158,7 @@ impl Value {
     /// For more information read [RFC6901](https://tools.ietf.org/html/rfc6901).
     pub fn pointer<'a>(&'a self, pointer: &str) -> Option<&'a Value> {
         fn parse_index(s: &str) -> Option<usize> {
-            if s.starts_with("+") || (s.starts_with("0") && s.len() != 1) {
+            if s.starts_with('+') || (s.starts_with('0') && s.len() != 1) {
                 return None
             }
             s.parse().ok()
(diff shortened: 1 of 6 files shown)


json/src/value.rs
```diff
@@ -468,10 +468,14 @@ struct WriterFormatter<'a, 'b: 'a> {
...
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        match self.inner.write_str(str::from_utf8(buf).unwrap()) {
-            Ok(_) => Ok(buf.len()),
-            Err(_) => Err(io::Error::last_os_error()),
+        fn io_error<E>(_: E) -> io::Error {
+            // Value does not matter because fmt::Debug and fmt::Display impls
+            // below just map it to fmt::Error
+            io::Error::new(io::ErrorKind::Other, "fmt error")
         }
+        let s = try!(str::from_utf8(buf).map_err(io_error));
```
oli-obk: nit: this could be `.or(io_error())` (+ remove the argument from `io_error`)

json/src/value.rs
```diff
@@ -468,10 +468,14 @@ struct WriterFormatter<'a, 'b: 'a> {
 
 impl<'a, 'b> io::Write for WriterFormatter<'a, 'b> {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        match self.inner.write_str(str::from_utf8(buf).unwrap()) {
-            Ok(_) => Ok(buf.len()),
-            Err(_) => Err(io::Error::last_os_error()),
```
oli-obk: the semantics changed, isn't the `last_os_error` important?

oli-obk:
do we need to change something in travis, too?

oli-obk:
Great cleanup. Can be merged as is if the comments are irrelevant.

json/src/value.rs
dtolnay: In `.or(io_error())` the argument is evaluated even if there is no error. And `io::Error::new` is [actually pretty expensive](https://github.com/rust-lang/rust/blob/de78655bca47cac8e783dbb563e7e5c25c1fae40/src/libstd/io/error.rs#L191-L204) - two memory allocations, one on the `into()` call that makes a Box, and another explicit `Box::new`. So this function would end up doing 4 useless memory allocations on every call to `write`... which may itself by called many many times as a value is being serialized.

json/src/value.rs
dtolnay: The only callers are the `fmt::Debug` and `fmt::Display` impls below, which immediately do `.map_err(|_| fmt::Error)` to throw away the error. I mention this in the comment inside `io_error`.

Also I doubt that `fmt::Formatter::write_str` or `str::from_utf8` would set `errno` anyway, right?

dtolnay:
> do we need to change something in travis, too?

No, I changed lib.rs so clippy runs in Travis now.

```diff
- #![cfg_attr(feature = "nightly-testing", plugin(clippy))]
+ #![cfg_attr(feature = "clippy", plugin(clippy))]
```

dtolnay:
Travis passes `--features nightly-testing` to json_test [here](https://github.com/serde-rs/json/blob/bfd835f2023ad0f2e789892f4151fac6ae7d236b/.travis.yml#L33) and json_test passes `--features clippy` to serde_json [here](https://github.com/serde-rs/json/blob/bfd835f2023ad0f2e789892f4151fac6ae7d236b/json_tests/Cargo.toml#L10).

PR #105, opened 2016-07-10
Specialize parse_string for Iter=&str

dtolnay:
When deserializing from a `&str`, we can assume that the input is valid UTF-8 so don't need to do additional checks.

This makes the [Twitter benchmark](https://github.com/serde-rs/json-benchmark) about 15% faster because it is heavy on strings.

### DIFF
--- a/json/src/read.rs
+++ b/json/src/read.rs
@@ -204,7 +204,8 @@ impl<'a> SliceRead<'a> {
                         &self.slice[start .. self.index]
                     } else {
                         scratch.extend_from_slice(&self.slice[start .. self.index]);
-                        scratch
+                        // "as &[u8]" is required for rustc 1.8.0
+                        scratch as &[u8]
                     };
                     self.index += 1;
                     return result(self, string);
--- a/json/src/read.rs
+++ b/json/src/read.rs
@@ -58,6 +58,11 @@ pub struct SliceRead<'a> {
     index: usize,
 }
 
+/// Elide UTF-8 checks by assuming that the input is valid UTF-8.
+pub struct StrRead<'a> {
+    delegate: SliceRead<'a>,
+}
+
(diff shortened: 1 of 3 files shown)


dtolnay:
Uh-oh, this does not compile on 1.8.0. Good thing I added a Travis build in #107.

```
src/read.rs:201:34: 208:22 error: if and else have incompatible types:
 expected `&[u8]`,
    found `&'s mut collections::vec::Vec<u8>`
(values differ in mutability) [E0308]
src/read.rs:201                     let string = if scratch.is_empty() {
src/read.rs:202                         // Fast path: return a slice of the raw JSON without any
src/read.rs:203                         // copying.
src/read.rs:204                         &self.slice[start .. self.index]
src/read.rs:205                     } else {
src/read.rs:206                         scratch.extend_from_slice(&self.slice[start .. self.index]);
```

dtolnay:
Fixed (and rebased so that the 1.8.0 build runs).

json/src/read.rs
```diff
@@ -173,6 +178,50 @@ impl<'a> SliceRead<'a> {
...
+    /// contains no backslash escape sequences, the returned &str is a slice of
+    /// the raw JSON data so we avoid copying into the scratch space.
+    fn parse_str_bytes<'s, T, F>(&'s mut self, scratch: &'s mut Vec<u8>, result: F) -> Result<T>
+        where T: 's,
+              F: FnOnce(&'s Self, &'s [u8]) -> Result<T>,
+    {
+        // Index of the first byte not yet copied into the scratch space.
+        let mut start = self.index;
+
+        loop {
```
oli-obk: "I've seen this code before" `*scroll*` "oh there it is" xD

oli-obk:
Awesome work

PR #110, opened 2016-07-11
Speed up float parsing

### LINKED ISSUE #108: Parsing floats is slow
These two lines are extremely expensive:

selection_003

We can entirely avoid the division and multiplication by using a lookup table, or some other better approach.

### LINKED ISSUE #109: Long integer causes the following byte to be lost
```rust
serde_json::from_str::<Vec<f64>>("[123456789012345678901]").unwrap();
```

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
Syntax("EOF while parsing a list", 1, 23)'
```

Here is the buggy code: [json/src/de.rs#L294-L302](https://github.com/serde-rs/json/blob/e059d03aea557bbd2c53ff5af18951c65da8f509/json/src/de.rs#L294-L302). The first one needs to be a peek.

dtolnay:
Fixes #108 and fixes #109.

Parse floats as a u64 until the end, then use a lookup table to scale by the correct power of 10. This is faster and also more accurate. The old code used repeated division by 10 which is so inaccurate that `1.0 / 10.0 / 10.0 / 10.0 / 10.0 / 10.0 / 10.0 != 1e-6`.

This is about 40% improvement on the [Canada benchmark](https://github.com/serde-rs/json-benchmark) (almost twice as fast).

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -2,9 +2,8 @@
 //!
 //! This module provides for JSON deserialization with the type `Deserializer`.
 
-use std::i32;
+use std::{f64, i32, i64, u64};
 use std::io;
-use std::str;
 use std::marker::PhantomData;
 
 use serde::de;
@@ -87,12 +86,9 @@ struct DeserializerImpl<R: Read> {
     str_buf: Vec<u8>,
 }
 
-macro_rules! try_or_invalid {
-    ($self_:expr, $e:expr) => {
-        match $e {
-            Some(v) => v,
-            None => { return Err($self_.error(ErrorCode::InvalidNumber)); }
-        }
+macro_rules! overflow {
+    ($a:ident * 10 + $b:ident, $c:expr) => {
+        $a >= $c / 10 && ($a > $c / 10 || $b > $c % 10)
     }
 }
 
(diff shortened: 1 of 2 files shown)


json/src/de.rs
```diff
@@ -404,35 +394,60 @@ impl<R: Read> DeserializerImpl<R> {
...
+            Some(pow) => {
+                if exponent >= 0 {
+                    significand as f64 * pow
+                } else {
+                    significand as f64 / pow
+                }
+            }
+            None => {
+                if exponent >= 0 && significand != 0 {
+                    f64::INFINITY
```
oli-obk: I don't think we should default to infinity. Not only does it break round tripping, it's also very confusing to get an infinity from a non-infinite number

json/src/de.rs
```diff
@@ -404,35 +394,60 @@ impl<R: Read> DeserializerImpl<R> {
...
+            c @ b'0' ... b'9' => { (c - b'0') as i32 }
             _ => { return Err(self.error(ErrorCode::InvalidNumber)); }
         };
 
         while let c @ b'0' ... b'9' = try!(self.peek_or_null()) {
             self.eat_char();
+            let digit = (c - b'0') as i32;
+
+            if overflow!(exp * 10 + digit, i32::MAX) {
+                // Really big exponent, just ignore the rest.
```
oli-obk: as with the other overflowing cases above, shouldn't we just bail out?

json/src/de.rs
dtolnay: This matches the current behavior (as of 0.7.4). I filed #112 to change the behavior.

```rust
assert_eq!(f64::INFINITY, serde_json::from_str("1e1000").unwrap());
```

json_tests/tests/test_json.rs
```diff
@@ -750,11 +750,14 @@ fn test_parse_number_errors() {
         (".", Error::Syntax(ErrorCode::ExpectedSomeValue, 1, 1)),
         ("-", Error::Syntax(ErrorCode::InvalidNumber, 1, 1)),
         ("00", Error::Syntax(ErrorCode::InvalidNumber, 1, 2)),
+        ("0x80", Error::Syntax(ErrorCode::TrailingCharacters, 1, 2)),
```
oli-obk: did those new tests fail before these changes? Or are they new failures

dtolnay: These failed before but were not tested.

PR #114, opened 2016-07-11
Error on large numbers instead of parsing infinity

### LINKED ISSUE #112: Error on large numbers instead of parsing infinity
Not only does it break round tripping, it's also very confusing to get an infinity from a non-infinite number.

dtolnay:
Fixes #112. Also gets us +5 points on the nativejson floating point conformance tests because of the support for large negative exponents.


dtolnay:
Looks like a Clippy false positive. I filed https://github.com/Manishearth/rust-clippy/issues/1086 and will push a workaround.

json/src/de.rs
```diff
@@ -424,29 +419,60 @@ impl<R: Read> DeserializerImpl<R> {
         self.visit_f64_from_parts(pos, significand, final_exp, visitor)
     }
 
+    #[inline(never)]
```
oli-obk: why `inline(never)`?

oli-obk:
lgtm. I'm assuming the `inline(never)` has a positive performance impact? Please note so with a comment.

PR #117, opened 2016-07-15
Run rustfmt

dtolnay:


dtolnay:
Not perfect, but much more consistent than what existed before.

Issue #126, opened 2016-07-30
API inconsistency, is_string vs as_str


shahn:
For 0.8.0, as_string and as_boolean were renamed as_str and as_bool ( aac2c26a9f26f0f69d78b7626b5c9548babf9d70 ). This breaking change is neither documented for the 0.8.0 release nor is it consistent with the existing is_boolean and is_string methods. It would probably be wise to restore consistency by changing the is_ functions as well, but I'm raising this here because maybe this was an unintended breaking change that should be rolled back?

dtolnay:
The as_* changes were intended as filed in #39. The breaking change *is* documented in the [serde_json 0.8.0 release notes](https://github.com/serde-rs/json/releases/tag/v0.8.0) (maybe you were looking only at the serde release notes).

> - Value::as_string() has been renamed to as_str() and Value::as_boolean() has been renamed to as_bool() to improve consistency (#92)

The is_* was an oversight. We will need to add is_str/is_bool in 0.8.1 and remove the old ones in 0.9.0.

shahn:
Hah, indeed I was looking for the serde_json release notes, didn't find them immediately and then assumed there were just serde release notes. My bad!

dtolnay:
I just noticed that [json-rust](http://terhix.com/doc/json/enum.JsonValue.html) does the same thing. The enum variant is called `JsonValue::String` so there is a method `is_string()` which checks if the JsonValue is the String variant. But `as_str()` gives you back a `&str`. For bool the discrepancy makes slightly more sense in their case because their variant is called Boolean (`is_boolean()`) while ours is called Bool.

dtolnay:
I guess it boils down to:

- whether variants should be named after what they store or what you can get out of them,
- whether `is_*` methods should be named after the variant or after the type you can get out of the variant,
- whether `as_*` methods should be named after the variant or after the type they return.

The API in 0.8.0 has [variants named after what they store](http://serde-rs.github.io/json/serde_json/value/enum.Value.html), `is_*` methods named after the variant, and `as_*` methods named after the type they return. I think all of those decisions make sense.

PR #127, opened 2016-07-30
Rename is_{string,boolean} to is_{str,bool}

shahn:
This matches the rename of as_{string,boolean} to as_{str,bool}. Fixes
issue #126

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -262,7 +262,7 @@ impl Value {
     }
 
     /// Returns true if the `Value` is a String. Returns false otherwise.
-    pub fn is_string(&self) -> bool {
+    pub fn is_str(&self) -> bool {
         self.as_str().is_some()
     }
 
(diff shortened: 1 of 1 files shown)


dtolnay:
We cannot remove is_string and is_boolean until the next breaking release. Please change the doc comments to indicate that these are deprecated and add is_str and is_bool as new functions.

oli-obk:
Are we on a new enough stable to support #[deprecated] ?

dtolnay:
Unfortunately not, we have one release to go. The best we can do is probably:

```rust
#[doc(hidden)]
#[cfg_attr(feature = "unstable",
           deprecated(since = "0.8.1",
                      note="will be removed in 0.9.0; use is_str() instead"))]
fn
```

shahn:
(Waiting on the discussion in #126 to update this PR)

dtolnay:
Thanks for the PR. Based on https://github.com/serde-rs/json/issues/126#issuecomment-236373219 there is a justification for the current names. Also they are consistent with json-rust which famously has a more extensive and better-designed Value API than us, so it can't be so bad.

Issue #135, opened 2016-08-09
Build fails for --features preserve_order 


laktak:
This fails at the moment:

```
$ cargo build --features preserve_order
   Compiling serde_json v0.8.0 (file:///Users/spark/prj/hjson/hjson-rust/_json/json)
src/value.rs:383:39: 383:48 error: no method named `serialize` found for type `&linked_hash_map::LinkedHashMap<std::string::String, value::Value>` in the current scope
src/value.rs:383             Value::Object(ref v) => v.serialize(serializer),
                                                       ^~~~~~~~~
src/value.rs:383:39: 383:48 note: the method `serialize` exists but the following trait bounds were not satisfied: `linked_hash_map::LinkedHashMap<std::string::String, value::Value> : serde::Serialize`
src/value.rs:467:53: 467:62 error: no method named `visit_map` found for type `linked_hash_map::serde::LinkedHashMapVisitor<_, _>` in the current scope
src/value.rs:467                 let values = try!(MapVisitor::new().visit_map(visitor));
                                                                     ^~~~~~~~~
src/value.rs:467:30: 467:72 note: in this expansion of try! (defined in <std macros>)
src/value.rs:467:53: 467:62 help: items from traits can only be used if the trait is in scope; the following trait is implemented but not in scope, perhaps add a `use` for it:
src/value.rs:467:53: 467:62 help: candidate #1: `use serde::de::Visitor`
error: aborting due to 2 previous errors
error: Could not compile `serde_json`.
```

Maybe also add it to travis?

laktak:
[troplin](https://users.rust-lang.org/t/stuck-with-trait-bounds-were-not-satisfied-not-in-scope/6861/2?u=laktak) pointed me in the right direction - linked_hash_map is still using serde 0.7 but no version referencing 0.8 has been released.

dtolnay:
I released serde_json 0.8.1 containing your fix.

PR #136, opened 2016-08-11
fix build for preserve_order

laktak:
fix for #135


laktak:
@dtolnay linked_hash_map does not build with Rust 1.8 - not sure how you would want to handle that.

dtolnay:
Thanks for the fix. Since 1.8 didn't work without this fix, we can ignore the failure.

Issue #128, opened 2016-08-02
Round trip floats


dtolnay:
Ideally this test would pass.

``` rust
extern crate serde_json;

#[macro_use]
extern crate quickcheck;

quickcheck! {
    fn floats(n: f64) -> bool {
        let j = serde_json::to_string(&n).unwrap();
        serde_json::from_str::<f64>(&j).unwrap() == n
    }
}
```

On the printing side grisu2 guarantees that the f64 closest to the string representation is identical to the original input, so the inaccuracy is somewhere on the parsing side.

Mark-Simulacrum:
I'm not sure the conclusion that the issue is with parsing is entirely correct (although there do seem to be problems in that area as well), since this test fails in the `dtoa` crate: `test_write(17.922589369123898, "17.922589369123898");`.
```
	thread 'test_f64' panicked at 'assertion failed: `(left == right)` (left: `"17.922589369123899"`, right: `"17.922589369123898"`)', tests/test.rs:36
```

dtolnay:
@Mark-Simulacrum the f64 that is closest to 17.922589369123899 is identical to the f64 that is closest to 17.922589369123898, so dtoa (grisu2) is correct here.

```rust
assert_eq!(17.922589369123898, 17.922589369123899);
```

As I wrote:

> grisu2 guarantees that the f64 closest to the string representation is identical to the original input

dtolnay:
In the extreme, you probably wouldn't expect this test to pass:

```rust
test_write(17.92258936912389800000000000000001f64, "17.92258936912389800000000000000001");
```

Mark-Simulacrum:
Hmm, I wonder how Javascript's JSON.parse/JSON.stringify deal with this. `JSON.stringify(JSON.parse('{ "test": 17.922589369123898 }'))` returns the expected (by me, at least) `17.922589369123898` despite `assert_eq!(17.922589369123898, 17.922589369123899);` being true in JS. Well, I somewhat expect that decode(encode(any f64)) == the same f64; but I guess that isn't the case at least for serde-json and json-rust: perhaps this isn't a good requirement.

dtolnay:
@Mark-Simulacrum using f64 you cannot have both of these return "the expected":

```js
JSON.stringify(JSON.parse('{ "test": 17.922589369123898 }'))
JSON.stringify(JSON.parse('{ "test": 17.922589369123899 }'))
```

because those are the same f64.

dtolnay:
[...]

PR #138, opened 2016-08-17
test value -> string -> value roundtripping with quickcheck

oli-obk:
so quickcheck found out 2 things:
1. roundtripping with floats is borked #128 
2. roundtripping with `Value::I64` will end up with a `Value::U64` if the value is positive #146


dtolnay:
Do you think there is value in taking out the I64/U64/F64 parts of this for now and merging the rest so at least we have quickcheck of everything else? We can add back the number stuff after we resolves those two issues.

dtolnay:
Let's revisit once #128 is implemented.

PR #140, opened 2016-08-19
Dont use Result for writes that can't fail.

bruno-medeiros:
Since writing to Vec can't fail, why not return successful value
direclty instead of wrapping it with a Result?


oli-obk:
It can actually fail at runtime, when you serialize e.g. a `HashMap<i32, T>`, because the key isn't a string.

bruno-medeiros:
@oli-obk  I see, interesting. Is that a limitation of JSON format, or is true for any serde format?

oli-obk:
just of JSON

PR #143, opened 2016-08-25
Extend Formatter with more callbacks

### LINKED ISSUE #142: Expand Formatter support
I want to offer pretty printing of JSON objects beyond what the `PrettyFormatter` can supply, for example color support or intelligent line wrapping.  For that, I'd like to expose more callbacks to control formatting.

Ideally I would like to be able to color every "token type" differently.

Currently, these things can be colored:

  * `{` `}` `[` `]` `,` `:` trivially in `fn open(...)`, `fn colon(...)`, `fn comma(...)` and `fn close(...)`
  * Field names (by checking if `ch == "{"` in `fn open(...)` and `fn comma(...)` and writing color info accordingly)
  * Field values and array elements (by writing color info in `fn open(...)`, `fn comma(...)` and `fn close(...)`)

Note that values in general cannot be colored because when serializing an `i32` for example, no formatting callbacks are called.  The current situation also leads to awkward code like this: https://github.com/dflemstr/rq/blob/bb845ea5318db377927cf10e75eff6d44194129c/src/value/json.rs#L142-L145

What would be nice is to have a callback for each JSON terminal.  This would involve:

  - [ ] Rename `fn open(...)` to `fn open_object/array(...)` (and `fn close(...)` to `fn close_object/array(...)`), but there can be a default method body for backwards compatibility.
  - [ ] Add callbacks for all missing JSON terminals:
    - [ ] `fn object_key(...)`
    - [ ] `fn string(...)`
    - [ ] `fn number(...)`
    - [ ] `fn bool(..., bool)`
    - [ ] `fn null(...)`
  - [ ] Maybe add support for low-level terminals:
    - [ ] `fn open/close_string(...)`.
    - [ ] `fn string_char(..., char)`
    - [ ] `fn string_escape(..., Escape)` where `enum Escape { Quote, ..., Unicode([u8; 4]) }` or something.
    - [ ] `fn number_digit(..., char)` etc.

This probably means that the `Serializer` would never write to its `Write` on its own, but delegate everything to the `Formatter`.

I could implement this change if people think it's a good idea.  I'm mostly seeking feedback on the approach first.

dflemstr:
Closes #142.  See that issue for discussion about the motivation; I would only like to discuss the code here.


dflemstr:
Hm, just realized I completely omitted the string handling code... will fix!

dtolnay:
Nicely done! I like this a lot. Just minor comments:

- Can you explain the reasoning for `Ok(())` as the default implementation for all of the Formatter methods? For most of those methods you would never want a Formatter implementing it as `Ok(())` so that doesn't make sense to me as a default. We could either not provide default methods, or provide the CompactFormatter's methods as the defaults (i.e. implement CompactFormatter as just `impl Formatter for CompactFormatter {}` with no overrides). I like the second because it codifies what the comments say - for example the comment says "this method is responsible for writing a `[`" and the implementation would write a `[`, the minimal implementation that satisfies the contract of the method. It would also remove the duplication between CompactFormatter and PrettyFormatter of all the null/bool/number/string methods.
- I think it's inconsistent to have `begin_number`/`end_number` but not `begin_bool`/`end_bool`. Could we remove `begin_number`/`end_number` and just have the Formatter call any begin/end logic from `write_integer`/`write_floating` if necessary?

json/src/ser.rs
```diff
@@ -56,6 +57,33 @@ impl<W, F> Serializer<W, F>
...
+        Ok(())
+    }
+
+    #[inline]
+    fn write_floating<V>(&mut self, value: V) -> Result<()>
+        where V: dtoa::Floating + num_traits::Float
+    {
+        match value.classify() {
+            FpCategory::Nan | FpCategory::Infinite => {
+                try!(self.writer.write_all(b"null"))
```
dtolnay: This should go through write_null.

dtolnay:
I repointed this PR to a new v0.9.0 branch. Once it gets merged I'll make an 0.9.0-alpha1 release so that you have something to depend on in your crate, but an actual 0.9.0 release may not happen for a while - nobody else is using this API so there wouldn't be a motivation for people to upgrade.

dflemstr:
@dtolnay I think this should do it.  Also, don't worry about a release; my crate is not on `crates.io` so I can depend on my git repo.

PR #149, opened 2016-09-12
take argument to `Value::to_value` by value

rphmeier:
This is backwards compatible as there exists an implementation of Serializable for `&'a T` where T: Serializable.

It makes the function a lot more friendly to use in iterator adaptors, option and result consumers, etc.


oli-obk:
> This is backwards compatible as there exists an implementation of Serializable for &'a T where T: Serializable.

I think so, too. Good thing we don't have HKT xD

rphmeier:
Yes, and although this also adds a `Sized` constraint I would say this is in fact more friendly to `!Sized` types as now any unsizing pointer can be passed.

dtolnay:
Thanks! I included this in the 0.8.2 release.

PR #150, opened 2016-09-14
Allow trait objects in the to_writer function.

withoutboats:
Write is an object safe trait, so users can construct an `&mut Write`
object. However, the `to_writer` and `to_writer_pretty` functions take an
`&mut W: Write`. In order to allow users to pass `&mut Write` objects to
these functions, their `W` parameter is now constrained `?Sized`, to allow
it to be a dynamically sized type (such as a trait object).

### DIFF
--- a/json/src/ser.rs
+++ b/json/src/ser.rs
@@ -950,7 +950,7 @@ fn fmt_f64_or_null<W>(wr: &mut W, value: f64) -> Result<()>
 
 /// Encode the specified struct into a json `[u8]` writer.
 #[inline]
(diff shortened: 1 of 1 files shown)


withoutboats:
An alternative solution is to take `W` by value, though I somewhat like that this API is very opinionated about how it will handle ownership, because that helps the user understand what this API will do.

dtolnay:
Looks good to me, and thanks for the explanation. I can't do it right now but tonight I will publish a release for you including this fix.

dtolnay:
I included this in the 0.8.2 release.

PR #153, opened 2016-09-16
Add as_borrow() generic method for Value

Kixunil:
This change adds as_borrow() generic method for Value. It allows users to use e.g. `some_fn(try!(value.as_borrow()))` instead of `some_fn(try!(value.as_string().ok_or(SomeCustomError)))`

Ability to borrow mutably may be even more interesting.

It also defines nice error type for situations when types don't match as well as representation of value type. Those things create dependency chain.

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -473,6 +473,99 @@ impl de::Deserialize for Value {
     }
 }
 
+/// Represents type of a JSON value
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum ValueType {
+    /// Represents a JSON null value type
+    Null,
+
+    /// Represents a JSON Boolean type
+    Bool,
+
+    /// Represents a JSON signed integer type
+    I64,
+
+    /// Represents a JSON unsigned integer type
+    U64,
+
+    /// Represents a JSON floating point number type
+    F64,
+
+    /// Represents a JSON string type
+    String,
+
+    /// Represents a JSON array type
+    Array,
+
+    /// Represents a JSON object type
+    Object,
+}
+
+impl ValueType {
+
+    /// Returns type of a value
+    /// 
+    /// # Example
+    ///
+    /// ```rust
+    /// extern crate serde_json;
+    ///
+    /// use serde_json::to_value;
+    /// use serde_json::value::ValueType;
+    /// use std::collections::BTreeMap;
+    ///
+    /// fn main() {
+    ///     let null = to_value(());
+    ///     let boolean = to_value(true);
+    ///     let int64 = to_value(-42i64);
+    ///     let uint64 = to_value(18446744073709551337u64);
+    ///     let float = to_value(3.1415926535f64);
+    ///     let string = to_value("Hello world!");
+    ///     let array = to_value(&[1, 2, 3, 4, 5]);
+    ///     let mut map = BTreeMap::new();
+    ///     map.insert("foo", to_value("bar"));
+    ///     let map = to_value(&map);
+    ///
+    ///     assert_eq!(ValueType::of(&null), ValueType::Null);
+    ///     assert_eq!(ValueType::of(&boolean), ValueType::Bool);
+    ///     assert_eq!(ValueType::of(&int64), ValueType::I64);
+    ///     assert_eq!(ValueType::of(&uint64), ValueType::U64);
+    ///     assert_eq!(ValueType::of(&float), ValueType::F64);
+    ///     assert_eq!(ValueType::of(&string), ValueType::String);
+    ///     assert_eq!(ValueType::of(&array), ValueType::Array);
+    ///     assert_eq!(ValueType::of(&map), ValueType::Object);
+    /// }
+    pub fn of(val: &Value) -> Self {
+        match *val {
+            Value::Null => ValueType::Null,
+            Value::Bool(_) => ValueType::Bool,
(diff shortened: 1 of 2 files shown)


json/src/value.rs
```diff
@@ -1333,3 +1512,139 @@ impl<T: ?Sized> ToJson for T
         to_value(&self)
     }
 }
+
+/// Just like `std::borrow::Borrow<Value>` but borrowing may fail.
+pub trait TryBorrow<'a>: 'a + ::std::marker::Sized {
+    /// Returns reference to internal data if type matches.
+    fn try_borrow(value: &'a Value) -> Result<Self, TypeError>;
```
oli-obk: this trait could include the `try_borrow_mut` method

json/src/value.rs
```diff
@@ -1333,3 +1512,139 @@ impl<T: ?Sized> ToJson for T
...
+}
+
+impl<'a> TryBorrow<'a> for &'a () {
+    fn try_borrow(value: &'a Value) -> Result<Self, TypeError> {
+        static EMPTY: () = ();
+        if value.is_null() { Ok(&EMPTY) } else { Err(TypeError { expected: ValueType::Null, provided: ValueType::of(value) }) }
+    }
+}
+
+impl<'a> TryBorrow<'a> for &'a bool {
```
oli-obk: all these impls should be abstracted away into a macro

json/src/value.rs
```diff
@@ -473,6 +559,99 @@ impl de::Deserialize for Value {
     }
 }
 
+/// Represents type of a JSON value
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum ValueType {
```
oli-obk: This type sounds a lot like [`std::mem::discriminant_value`](https://github.com/rust-lang/rust/pull/34785), so we could just wait until that is stable (3 Months probably)

Kixunil: I agree. The library is still unstable, so we can wait. I hope time is not concern to anyone else.

I think I'm going to leave this as is for now until discriminant is stabilized. However, I'm not sure whether discriminant will allow to do exactly what I mean. I will check that out.

json/src/value.rs
Kixunil: The reason is to mimic `Borrow` and `BorrowMut`. I'm not even sure whether it would be possible to have them together in one trait, since one is defined for `&'a T` and the other for `&'a mut T`...

However, I'm open to put them together, if it's possible and advantageous.

json/src/value.rs
Kixunil: I was thinking about it too lately. I'm not sure how to define the macro to be readable enough. I was thinking about something like this:

`impl_try_borrows!(
   Bool => bool,
   I64 => i64,
// etc.
);`

What do you think?

Kixunil:
The motivation is mostly to let inference figure out the type. It would allow users to assign the resulting borrow into struct or passing it to function without thinking about types. Also, if they later decide to change the type, it will adapt automatically.

I'm not sure whether this will lead to unreadable messages. Do you have an example of such message?

I also noticed issue #126 (see also #127). This PR solves the naming problem quite elegantly. Instead of specifying the name in function (and deciding which is the best), it allows user to specify the exact type. At least for `.as_*` (Personally, I'd drop all `.as_*` fns and replace it with this one. Possibly also with moving counterparts.) This is especially useful in case of `&str`/`&String` or even more in mutable cases.

I believe something similar could be implemented for `.is_*` case but that could be also viewed as complicated. That's why I didn't code it yet. I'm willing to write the change if it's desired.

Finally, please note that I intentionally used custom error type, so it can be displayed to user in very human friendly way. ("type mismatch: expected type: X found type: Y")

I hope I made myself clear enough. I'm looking forward for your thoughts!

dtolnay:
@oli-obk what is the status of this PR? Is it waiting for stable `std::mem::discriminant`? Why is that better than `ValueType`? Have you found any unreadable error messages from this?

json/src/value.rs
oli-obk: Sounds good to me

json/src/value.rs
oli-obk: Right, I forgot about that

oli-obk:
> Why is that better than ValueType

Just less code for the same effect

> Have you found any unreadable error messages from this?

Imagine calling the method on a `&str` target. There's no impl for it and no way for the user to write one

Will review more later

dtolnay:
Reading through this again, I don't think this would be a beneficial change. It seems like the three main motivations for this PR are:

- Genericity, being able to write a function that takes a type parameter controlling what type to borrow as. Is this a real use case?
- Flexibility, being able to borrow as &str as well as &String. Can you give any examples of real code that is not well served by the existing API? I think all of the examples in the doc comments would be as clear or clearer using the existing methods.
- Error messages, like "expected type: Map found type: String". I think this PR would actually be detrimental to error messages. If my app crashes with that message, I have no idea what to do. In Serde 0.9 we have tried to hold a somewhat higher bar for error messages than in previous releases, and sometimes that means forcing people to think about what error case they are handling and how. An actionable message requires more context than what would be available in Value::as_borrow, for example "book.json: expected configuration for \`edit-link\` plugin to be a map"

Kixunil:
This PR was posted a long time ago and serde changed quite a lot since then. I'd have to look at it again and re-validate. I'm fine with closing now.

PR #156, opened 2016-09-25
Ensure value::Serializer emits Value::Null for non-finite floats

### LINKED ISSUE #155: to_value(NAN) should be null
```rust
extern crate serde_json;

fn main() {
    let value = serde_json::to_value(::std::f64::NAN);
    println!("value: {}", value);
    println!("value.is_null(): {}", value.is_null());
}
```

Expected:

```
value: null
value.is_null(): true
```

Actual:

```
value: null
value.is_null(): false
```

traviskaufman:
See https://tc39.github.io/ecma262/#sec-serializejsonproperty
Fixes #155 

--

As a heads up, I'm pretty new to rust so if I'm off-track with this PR I apologize! I ran the tests using the nightly compiler (as that was the only one that seemed to allow the given compiler features).

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -643,7 +643,11 @@ impl ser::Serializer for Serializer {
 
     #[inline]
     fn serialize_f64(&mut self, value: f64) -> Result<(), Error> {
-        self.value = Value::F64(value);
+        self.value = if value.is_finite() {
+            Value::F64(value)
+        } else {
+            Value::Null
+        };
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


dtolnay:
Looks good otherwise. Thanks for the fix!

json/src/value.rs
```diff
@@ -643,7 +643,10 @@ impl ser::Serializer for Serializer {
 
     #[inline]
     fn serialize_f64(&mut self, value: f64) -> Result<(), Error> {
-        self.value = Value::F64(value);
+        self.value = match value {
```
traviskaufman: Thanks for the tip! Will change.

traviskaufman:
changes made!

PR #161, opened 2016-10-15
Typo fixed: it's called Bool, not Boolean

mephinet:
cf https://docs.serde.rs/serde_json/value/enum.Value.html


dtolnay:
Thanks!

Issue #162, opened 2016-10-26
Add configurable recursion limit


erickt:
We've talked about this before, but apparently we don't have a ticket. We should add a configurable depth counter to allow deserializers to limit how deep of a structure they can deserialize. This would help to avoid stack overflows from maliciously constructed objects.

frewsxcv:
Relevant issue: https://github.com/serde-rs/serde/issues/82

dtolnay:
I added a recursion limit of 128 in #163. Are there any use cases that require this to be configurable?

dtolnay:
Nobody has complained.

himat:
Hi I would like to possibly complain.

I see that an option to disable the recursion limit was added https://docs.rs/serde_json/1.0.59/serde_json/struct.Deserializer.html#method.disable_recursion_limit 

But for my use case, I still want to make sure that there won't be some hugely nested JSON that I would need to deserialize and so I still want a max recursion limit, just one that is higher than the current default.
If I use that `disable_recursion_limit` function, how would I go about still ensuring that I have a max limit?

obi1kenobi:
https://github.com/obi1kenobi/cargo-semver-check/issues/108 would also very much benefit from a configurable limit.

In particular, a configurable limit is *much* more palatable than any of the other alternative approaches I can see, listed here: https://github.com/obi1kenobi/cargo-semver-check/issues/108#issuecomment-1232182385

wrwg:
It would be really nice to provide better protection from serialization attacks by allowing the nesting to be defined on a per-type level. Consider e.g. this example:

```
#[derive(Serialize, Deserialize)]
struct SystemMessage {
   user_value: SomeUserStruct
 }
 ```
 
Now consider some attacker constructs a `user_value` exactly at the allowed nesting_depth. When you try to serialize this value embedded in the other value, serialization fails, even though the user value by itself is valid and has been, say, obtained by deserialization. You really like to ensure that SystemMessage has a deeper nesting level than SomeUserStruct.

Also, one can imagine type layouts and encodings where a too deeply nested recursive type leads to a "decompression bomb". That is you have a relative small binary encoding, but if its deserialized it expands to something much larger leading to OOM. In such cases, you'd like to restrict the nesting depth to a significant smaller value than the current default.

SichangHe:
I am now complaining, @dtolnay, because the error message confused the heck out of me. And, it was hard for me to find out what exactly was the cause of the error.

[...]

PR #163, opened 2016-10-27
Limit recursion to 128 levels

dtolnay:


dtolnay:
I tried a few ways of factoring this into a helper but they were all slower in benchmarks. 701d84ef30b773ff820a2dee727e69a9a00b89a1 is ~3% slower. As is, this PR has no effect on benchmarks.

oli-obk:
Did you try slapping inline-always on the function?

dtolnay:
I tried `#[inline(always)]` on that helper, as well as `#[inline(always)]` with each of these three variations factored out of the helper:

```rust
#[cold]
#[inline(never)]
fn stack_overflow_error<T>(&mut self) -> Result<T> {
    Err(self.peek_error(ErrorCode::Custom("recursion limit exceeded".into())))
}
```

```rust
#[cold]
#[inline(never)]
fn stack_overflow_error(&mut self) -> Error {
    self.peek_error(ErrorCode::Custom("recursion limit exceeded".into()))
}
```

```rust
#[cold]
#[inline(never)]
fn stack_overflow_error() -> ErrorCode {
    ErrorCode::Custom("recursion limit exceeded".into())
}
```

Those range from 3% (the last one) to 10% (the middle one) slower than this PR.

10% is pretty astonishing given that we are talking about canada.json which is full of floating point numbers and parsing those is slow to begin with, so something weird is going on. I will check it out in `perf` when I have time.

dtolnay:
I get pretty much the same slowdowns on my laptop Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz and my desktop Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz.

json/src/de.rs
```diff
@@ -107,6 +108,7 @@ impl<R: Read> DeserializerImpl<R> {
         DeserializerImpl {
             read: read,
             str_buf: Vec::with_capacity(128),
+            remaining_depth: 128,
```
erickt: How about adding a `Deserializer::with_recursion_limit` method so this can be customized?  Or maybe a `Deserializer` builder with a `with_recursion_limit` method?

json/src/de.rs
```diff
@@ -205,12 +207,30 @@ impl<R: Read> DeserializerImpl<R> {
                 visitor.visit_str(s)
             }
             b'[' => {
+                self.remaining_depth -= 1;
```
erickt: I suppose all this could be moved into the `SeqVisitor` but I'm not sure if that'd make things any more clean.

json/src/de.rs
```diff
@@ -523,6 +543,10 @@ impl<R: Read> DeserializerImpl<R> {
     }
 }
 
+fn stack_overflow() -> ErrorCode {
+    ErrorCode::Custom("recursion limit exceeded".into())
```
erickt: It's a shame we can't add error codes in a backwards compatible fashion. Next major release though :)

dtolnay:
I still can't make a better factored version run as fast. I will give it another shot once we have likely/unlikely https://github.com/rust-lang/rust/issues/26179.

json/src/de.rs
dtolnay: I am keeping #162 open to decide whether this needs to be configurable.

Issue #68, opened 2016-05-13
add pointer_mut


oli-obk:
> It should be possible to add a pointer_mut(&str) method to Value. This would allow to add and modify values. (Maybe even a delete method) I failed to add such a method because of borrow checker.

https://github.com/serde-rs/json/pull/41#issue-136929480

dtolnay:
Added in #171 and released in 0.8.4.

PR #171, opened 2016-12-03
Implement issue #68 - add pointer_mut

CryptArchy:
I was thinking that JSON Patch support would be useful, and saw the Issues for that and for `pointer_mut` support. #68 `pointer_mut` seemed necessary for implementing Patch, so I took a swing at it! This is my first open-source Rust PR so criticism is welcome but be gentle, haha!

I moved `pointer` into a new trait (`Pointer`) and added `pointer_mut` to it as well.  To maintain symmetry I also added a `pointer_owned` method that consumes the `Value` and returns an owned `Value`. Because of the new trait this is a breaking change (you need to import the trait for `pointer` to work now.)

Unit tests were added and all pass, documentation was added but is mostly just copy-paste from the existing code. I also wrote a benchmark to verify that the changes I made to `pointer` weren't slower than the existing code, but left out the change to the `.rs.in` file I made to run the bench. On my machine, the new and old `pointer` methods were very close in performance. The new one actually seems slightly faster but they were within deviation range of each other so I'm making no claims except that my refactors shouldn't have negatively affected the method.

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -97,6 +97,13 @@ pub enum Value {
     Object(Map<String, Value>),
 }
 
+fn parse_index(s: &str) -> Option<usize> {
+    if s.starts_with('+') || (s.starts_with('0') && s.len() != 1) {
+        return None;
+    }
+    s.parse().ok()
+}
+
 impl Value {
     /// If the `Value` is an Object, returns the value associated with the provided key.
     /// Otherwise, returns None.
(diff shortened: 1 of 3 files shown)


oli-obk:
Why did you move everything into a trait? Do you have a use case for implementing those methods on another type?

Don't worry about the failing travis, it's just nightly and not related to serde directly.

Do you think it would be feasible to implement those methods through a macro so there's only one impl written?

json/src/pointer.rs
```diff
@@ -0,0 +1,141 @@
...
+            return None;
+        }
+        let tokens = pointer.split('/').skip(1).map(|x| x.replace("~1", "/").replace("~0", "~"));
+        let mut target = self;
+
+        for token in tokens {
+            let target_opt = match target {
+                Value::Object(mut map) => map.remove(&token[..]),
+                Value::Array(mut list) => {
+                    parse_index(&token[..]).and_then(|x| Some(list.remove(x)))
```
oli-obk: This should probably be `map` instead of and_then

CryptArchy:
Corrected the `map` instead of `and_then`.  Good catch, thank you!

I was using a trait to keep things clean while I worked, but now that I think about it more it does seem odd. If made generic over the return value (instead of being `serde_json::Value`) then the trait could generically represent any means of accessing elements via strings, like for XPath maybe.  But I acknowledge that's stretching things a bit.

It's no problem for me to move it out of the trait and put the methods into the Value impl as before.  

I'm not sure I understand what you mean about a macro.  Like doing a `pointer!(v: Value, s: &str)` kind of thing and have the macro be generic in the return based on the ref-type of the parent Value?

json/src/pointer.rs
```diff
@@ -0,0 +1,141 @@
+//! JSON Pointer
+//!
+//! This module presents JSON Pointers through the `Pointer` trait and implements it for the `Value` type.
+//!
+//! For more information read [RFC6901](https://tools.ietf.org/html/rfc6901).
+
+use value::Value;
+
+/// Provides the `pointer` method for locating values within an object using a string path.
+pub trait Pointer {
```
oli-obk: Yes, please move these methods directly onto the type. If we ever get a more concrete use case for splitting them off, we'll do that then

json/src/pointer.rs
```diff
@@ -0,0 +1,141 @@
...
+        let tokens = pointer.split('/').skip(1).map(|x| x.replace("~1", "/").replace("~0", "~"));
+        let mut target = self;
+
+        for token in tokens {
+            let tgt = target;
+            let target_opt = match *tgt {
+                Value::Object(ref mut map) => map.get_mut(&token[..]),
+                Value::Array(ref mut list) => {
+                    if let Some(idx) = parse_index(&token[..]) {
+                        list.get_mut(idx)
```
oli-obk: For symmetry with the other methods, also use `map` here

json/src/pointer.rs
```diff
@@ -0,0 +1,141 @@
...
+            return None;
+        }
+        let tokens = pointer.split('/').skip(1).map(|x| x.replace("~1", "/").replace("~0", "~"));
+        let mut target = self;
+
+        for token in tokens {
+            let target_opt = match *target {
+                Value::Object(ref map) => map.get(&token[..]),
+                Value::Array(ref list) => {
+                    parse_index(&token[..]).and_then(|x| list.get(x))
```
oli-obk: I know this was in there before, but can you check whether it works to just pass `token` instead of `&token[..]`

json/src/pointer.rs
CryptArchy: Done.

json/src/pointer.rs
CryptArchy: It won't take `token` but it will take just `&token`.  I wasn't sure why the full range slice was in there, I just assumed there was some subtlety I was missing.

json/src/pointer.rs
CryptArchy: Done.  I had originally had trouble with the closure, but it wasn't hard to correct (just needs a `move`) once I worked out the rest of the borrow checker's nagging. 😃

CryptArchy:
Ah ok, I originally thought you had meant something like that, and I had considered it.  At the moment, my brain isn't making the leap to how I would write it (or maybe I'm just not very good at macros yet!) The code does seem so very similar that it feels like some abstraction should work there, but the changes for each function are subtle and pervasive so it'd be tricky.

json/src/value.rs
```diff
@@ -161,26 +168,97 @@ impl Value {
...
+    ///
+    /// A Pointer is a Unicode string with the reference tokens separated by `/`.
+    /// Inside tokens `/` is replaced by `~1` and `~` is replaced by `~0`. The
+    /// addressed value is returned and if there is no such value `None` is
+    /// returned.
+    ///
+    /// For more information read [RFC6901](https://tools.ietf.org/html/rfc6901).
+    pub fn pointer_owned(self, pointer: &str) -> Option<Value> {
+        if pointer == "" {
+            // except for this, the function could probably be `&mut self` and only consume the child element
```
oli-obk: that is a good point!

@dtolnay what do you think about using `&mut self` and turning it into `Value::Null` in the case of `pointer == ""`?

dtolnay: I would prefer to keep it the way @CryptArchy has written it. The behavior you suggested is `v.pointer_mut(p).map(|m| mem::replace(m, Value::Null))` which makes it clear in the caller's code exactly how the value is being mutated.

dtolnay:
Please remove the carriage return characters from test_json.rs. That is why the file looks like it has 3396 lines of diff.

CryptArchy:
Oh, just noticed that. I had to fix a merge conflict and the diff tool apparently loves to mess up line endings. Sorry about that.

dtolnay:
:+1: Looks good to me. I'll give @oli-obk the final say because he has been more involved in this PR.

CryptArchy:
So I tested out what @dtolnay was saying and it looks like his little one liner is enough to completely replace the `point_owned` behavior.  As in:

```rust
pub fn pointer_owned(&mut self, pointer: &str) -> Option<Value> {
        use std::mem;
        self.pointer_mut(pointer).map(|m| mem::replace(m, Value::Null))
}
```

So that's pretty cool! I forget about the `mem` library, should probably use it more often. Should I push that version up, or since it's so simple do we want to even keep the `pointer_owned` version anymore?

CryptArchy:
As an aside, the `pointer_owned` method I said was for symmetry, but it was actually an experiment.  I implemented it first to better understand the ownership behavior around the Value graph, as it was a lot easier than the mutable case. Then I just thought it seemed useful to leave in, hehe.

CryptArchy:
Sorry, lots of babbling on my part. @oli-obk if you want me to change or remove `pointer_owned` then I'm fine either way.  Or if you like the current implementation then it can stay as is, of course. Up to you, just let me know.

oli-obk:
Yea, I'd say we remove it without an explicit use case, because I think it's problematic to move out the entire value object just to extract a single thing by value. All non-nulling designs I can think of return both the value and the "rest" (if there is any rest), but none are ergonomic.

CryptArchy:
Alright, removed `pointer_owned` and added a little more tests and documentation to cover David's neat little `mem::replace` trick and otherwise show how to use `pointer_mut`.

oli-obk:
Wonderful, thanks for being so thorough

CryptArchy:
My pleasure, thanks for the code reviews and being patient while I fixed things up. Looking forward to the next one, haha.

PR #172, opened 2016-12-05
Allow map keys that are effectively strings

dtolnay:
Fixes https://github.com/Aaronepower/tokei/issues/83. cc @Aaronepower

### DIFF
--- a/json/src/ser.rs
+++ b/json/src/ser.rs
@@ -456,6 +456,27 @@ impl<'a, W, F> ser::Serializer for MapKeySerializer<'a, W, F>
         self.ser.serialize_str(value)
     }
 
+    #[inline]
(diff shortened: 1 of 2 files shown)


oli-obk:
If we are serializing unit variants, shouldn't we also serialize unit structs and units ( the latter as the empty string)? Does roundtripping work/ should it?

dtolnay:
Unit structs and units are different because they serialize as `null` which is not a string. Unit variants serialize as `"VariantName"` which is a string. I don't think it is obvious to expect `null` in a map key to behave like an empty string (I know, even though we deserialize `null` to `String` as an empty string). Also who uses maps with unit keys?

Round tripping works, I added a test.

Issue #45, opened 2016-02-27
Consider serializing map integer keys as strings


erickt:
Right now serde_json rejects serializing maps with integer keys because semantically speaking, JSON only supports maps with string keys. There are workarounds in serde 0.7 with the new `#[serde(serialize_with="...", deserialize_with="...)]` (see this [gist](https://gist.github.com/erickt/535f3867c70db235bad7)), but it still can be annoying if this keeps causing problems.

Is there any real value about erroring out on non-key values?

ArtemGr:
I stumbled into the "key must be a string" error while using the newtype pattern (https://aturon.github.io/features/types/newtype.html).

```
pub struct Kind (pub String);
BTreeMap<Kind, Json>
```

Also I often use the small string optimization (around https://github.com/fitzgen/inlinable_string) to avoid the heap allocations. I gather Serde won't let me do `BTreeMap<InlString, Json>` because it thinks that `InlString` is not a string.

oli-obk:
@ArtemGr it shouldn't be a problem, you just need to make sure that your `Serialize` impl calls `serialize_str`. `InlString` is probably serialized as a struct.

wrt newtypes, I'm not sure if it's a good idea to make the default serialization work in this case. You created a newtype out of a reason (type safety), so I'm not sure if serde should break that as the default. But I'm open to be convinced otherwise ;)

ArtemGr:
Oh, thanks for the hint @oli-obk. I thought the problem is the string check but it looks like it's the lack of newtype support.

I sure would like for Serde to support the newtype pattern, maybe with some kind of new tag.

oli-obk:
serde does support newtypes. It's just a question of whether it would be a good default to serialize newtypes exactly like their inner type.

ArtemGr:
Yeah, what I mean is the transparent newtype support.

When we write `BTreeMap<String, u32>` we add type information that can't be represented in JSON: that the value is an unsigned integer and only 32 bits long. That's all well and good, because Rust is a more statically typed language than JavaScript and because serializing the type information is costly and unnecessary.

But with newtype pattern it suddenly works differently, Serde wants to actually write down in JSON that `struct Foo (String)` is a Foo string, not just any string.

> whether it would be a good default

I guess defaulting transparent newtype support would be a breaking change. It would be much easier to just stick a new tag on it.

oli-obk:
>  It would be much easier to just stick a new tag on it.

[...]

PR #177, opened 2016-12-24
Allow integers to be used as map keys

bjorn3:


json/src/ser.rs
```diff
@@ -489,44 +489,44 @@ impl<'a, W, F> ser::Serializer for MapKeySerializer<'a, W, F>
         Err(Error::Syntax(ErrorCode::KeyMustBeAString, 0, 0))
     }
 
-    fn serialize_isize(&mut self, _value: isize) -> Result<()> {
-        Err(Error::Syntax(ErrorCode::KeyMustBeAString, 0, 0))
+    fn serialize_isize(&mut self, value: isize) -> Result<()> {
+        self.ser.serialize_str(&value.to_string())
```
oli-obk: That allocates on the heap, can you find a way around that? Like write a `"`, the serialize normally, write another quote?

bjorn3:
Added a test and it doesn't heap allocate anymore.

dtolnay:
> This has probably been discussed before... I'm not sure we want this, and if we do, we need deserialization along with it

We do want this. The discussion is in #45. Do you have any reasons against?

json_tests/tests/test_json.rs
```diff
@@ -1501,10 +1501,24 @@ fn test_deserialize_from_stream() {
...
     let map = treemap!(
-        1 => 2,
-        3 => 4
+        true => 2,
+        false => 4
+    );
+
+    match serde_json::to_vec(&map).unwrap_err() {
+        serde_json::Error::Syntax(serde_json::ErrorCode::KeyMustBeAString, 0, 0) => {}
+        _ => panic!("integers used as keys"),
```
dtolnay: Please update this error message

json_tests/tests/test_json.rs
```diff
@@ -1501,10 +1501,24 @@ fn test_deserialize_from_stream() {
...
+    }
+}
+
+#[test]
+fn test_serialize_rejects_adt_keys() {
```
dtolnay: Please update the panic message below - it still refers to integers

json_tests/tests/test_json.rs
```diff
@@ -1723,3 +1737,26 @@ fn test_stack_overflow() {
...
+#[test]
+fn test_allow_ser_integers_as_map_keys(){
+    let map = treemap!(
+        1 => 2,
+        2 => 4,
+        -1 => 6,
+        -2 => 8
+    );
+    
+    serde_json::to_vec(&map).unwrap();
```
dtolnay: Please test the actual JSON string that it produces. You might be surprised what it is right now:

```json
{"\""-2"\"":8,"\""-1"\"":6,"\""1"\"":2,"\""2"\"":4}
```

json_tests/tests/test_json.rs
```diff
@@ -1723,3 +1737,26 @@ fn test_stack_overflow() {
...
+}
+
+/*#[test]
+fn test_allow_de_integers_as_map_keys(){
```
dtolnay: Are you interested in implementing this yourself in this PR or in another PR, or would you prefer to leave it to somebody else?

json/src/ser.rs
```diff
@@ -489,44 +489,48 @@ impl<'a, W, F> ser::Serializer for MapKeySerializer<'a, W, F>
...
 
-    fn serialize_i32(&mut self, _value: i32) -> Result<()> {
-        Err(Error::Syntax(ErrorCode::KeyMustBeAString, 0, 0))
+    fn serialize_i32(&mut self, value: i32) -> Result<()> {
+        self.serialize_i64(value as i64)
     }
 
-    fn serialize_i64(&mut self, _value: i64) -> Result<()> {
-        Err(Error::Syntax(ErrorCode::KeyMustBeAString, 0, 0))
+    fn serialize_i64(&mut self, value: i64) -> Result<()> {
```
dtolnay: I think this should be:

```rust
try!(self.ser.writer.write_all(b"\""));
try!(self.ser.serialize_i64(value));
try!(self.ser.writer.write_all(b"\""));
Ok(())
```

The `serialize_str` method writes a double-quoted JSON string. We just need to write a byte directly into the output.

oli-obk:
> Do you have any reasons against?

No, I want this, but I didn't remember that issue or what became of the discussion. Let's do it

json/src/ser.rs
bjorn3: Done

json_tests/tests/test_json.rs
bjorn3: Will try it in another PR.

json_tests/tests/test_json.rs
oli-obk: You missed thismin your update

json_tests/tests/test_json.rs
oli-obk: Same here

json_tests/tests/test_json.rs
oli-obk: And here

bjorn3:
Pushed a commit to improve build times on travis. Feel free to omit it.

oli-obk:
Hmm github doesn't allow me to approve @dtolnay's requested changes

Everything looks good to me now. Also, good idea with the cargo cache. I wonder if that will go into crazy overdrive someday with incremental compilation.

json_tests/tests/test_json.rs
bjorn3: Created PR serde-rs/serde#655

arthurprs:
Did we regress this between 0.9 and 1.0? I'm still investigating but I have some code choking on deserializing a struct containing a map<int,...>

dtolnay:
Yes, I'm working on a fix.

PR #178, opened 2016-12-24
Require newline to be written successfully

dtolnay:
Maybe we should run unit tests against an io::Write that returns Ok(0) on every other call...


oli-obk:
If we define the random sequence of Ok(0) as part of the input we can quickcheck it

oli-obk:
Also we really need a clippy lint for ignored return value

dtolnay:
I don't think randomness or quickcheck would be advantageous here. A writer that alternates between Ok(0) and Ok(len) on every other call would deterministically catch all mistakes like this.

PR #185, opened 2017-01-13
Update to Serde 0.9.0-rc1

dtolnay:


dtolnay:
I confirmed that so far there is zero performance difference compared to the most recent 0.8 release.

dtolnay:
Once we finish https://github.com/serde-rs/serde/pull/686 and https://github.com/serde-rs/serde/pull/687 I will make a 0.9.0-rc1 release for this PR to depend on.

PR #188, opened 2017-01-15
Extract serde_json::Map and serde_json::Number opaque newtypes

### LINKED ISSUE #57: Unify the number variants in Value

### LINKED ISSUE #152: Feature `preserve_order` is not additive
Cargo unions all features that all rdeps in a project of a crate want, and compiles the crate with that. This requires features to be additive. Since this feature changes the types used in the public API of the crate, it is not.

dtolnay:
Fixes #152 and fixes #57.

### DIFF
--- a/json/src/map.rs
+++ b/json/src/map.rs
@@ -1,0 +1,305 @@
+use serde::{ser, de};
+use std::fmt::{self, Debug};
+use value::Value;
+use std::hash::Hash;
+use std::borrow::Borrow;
+
+#[cfg(not(feature = "preserve_order"))]
+use std::collections::{BTreeMap, btree_map};
+
+#[cfg(feature = "preserve_order")]
+use linked_hash_map::{self, LinkedHashMap};
+
+/// Represents a key/value type.
+pub struct Map<K, V>(MapImpl<K, V>);
+
+#[cfg(not(feature = "preserve_order"))]
+type MapImpl<K, V> = BTreeMap<K, V>;
+#[cfg(feature = "preserve_order")]
+type MapImpl<K, V> = LinkedHashMap<K, V>;
+
+impl Map<String, Value> {
+    /// Makes a new empty Map.
+    pub fn new() -> Self {
+        Map(MapImpl::new())
+    }
+
+    #[cfg(not(feature = "preserve_order"))]
+    /// Makes a new empty Map with the given initial capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        let _ = capacity;
+        Map(BTreeMap::new()) // does not support with_capacity
+    }
+
+    #[cfg(feature = "preserve_order")]
+    /// Makes a new empty Map with the given initial capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        Map(LinkedHashMap::with_capacity(capacity))
+    }
+
+    /// Clears the map, removing all values.
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    /// Returns a reference to the value corresponding to the key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but the ordering
+    /// on the borrowed form *must* match the ordering on the key type.
+    pub fn get<Q>(&self, key: &Q) -> Option<&Value>
+        where String: Borrow<Q>,
+              Q: Ord + Eq + Hash + ?Sized
+    {
+        self.0.get(key)
+    }
+
+    /// Returns true if the map contains a value for the specified key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but the ordering
+    /// on the borrowed form *must* match the ordering on the key type.
+    pub fn contains_key<Q>(&self, key: &Q) -> bool
+        where String: Borrow<Q>,
+              Q: Ord + Eq + Hash + ?Sized
+    {
+        self.0.contains_key(key)
+    }
+
+    /// Returns a mutable reference to the value corresponding to the key.
+    ///
+    /// The key may be any borrowed form of the map's key type, but the ordering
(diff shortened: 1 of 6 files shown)


dtolnay:
I added #57 to this as well. Confirmed no performance difference once everything has `#[inline]` (it was about 20% without :astonished:).

json/src/map.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    /// Gets an iterator over the values of the map.
+    #[inline]
+    pub fn values(&self) -> MapValues {
+        MapValues(self.0.values())
+    }
+}
+
+impl Default for Map<String, Value> {
```
oli-obk: can't we just derive `Default`, `Clone`, `PartialEq` and `Debug` (although the latter might be better with the custom impl, to not generate `Map` in the output)

json/src/map.rs
```diff
@@ -0,0 +1,335 @@
...
+        self.0.fmt(formatter)
+    }
+}
+
+impl ser::Serialize for Map<String, Value> {
+    #[inline]
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where S: ser::Serializer
+    {
+        use serde::ser::SerializeMap;
```
oli-obk: I think we could just call `Serialize::serialize(&self.0, serializer)`.

json/src/map.rs
```diff
@@ -0,0 +1,335 @@
...
+        map.end()
+    }
+}
+
+impl de::Deserialize for Map<String, Value> {
+    #[inline]
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where D: de::Deserializer
+    {
+        struct Visitor;
```
oli-obk: also use forwarding here.

json/src/number.rs
```diff
@@ -0,0 +1,201 @@
...
+
+/// Represents a JSON number, whether integer or floating point.
+#[derive(Clone, PartialEq)]
+pub struct Number(N);
+
+// "N" is a prefix of "NegInt"... this is a false positive.
+// https://github.com/Manishearth/rust-clippy/issues/1241
+#[cfg_attr(feature = "clippy", allow(pub_enum_variant_names))]
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum N {
```
oli-obk: no need for this to be public

json/src/value.rs
```diff
@@ -707,7 +681,7 @@ impl ser::Serializer for Serializer {
     }
 
     fn serialize_bytes(self, value: &[u8]) -> Result<Value, Error> {
-        let vec = value.iter().map(|b| Value::U64(*b as u64)).collect();
+        let vec = value.iter().map(|&b: &u8| Value::Number(b.into())).collect();
```
oli-obk: full type annotation is not required afaik: https://is.gd/E0gcgl

json/src/map.rs
dtolnay: Except I want to stop relying on linked-hash-map's Serde impls. Right now this code works but your code wouldn't because linked-hash-map has serde 0.8 impls and not 0.9 impls.

json/src/map.rs
dtolnay: Same reason - linked-hash-map's impl still uses MapVisitor::end which doesn't exist in 0.9.

json/src/number.rs
dtolnay: Good catch, I was originally matching on this in value.rs.

json/src/value.rs
dtolnay: Done.

json/src/map.rs
dtolnay: The derived impls don't work:

- Ord is required for BTreeMap's Default:

    ```
    error[E0277]: the trait bound `K: core::cmp::Ord` is not satisfied
      --> src/map.rs:15:22
       |
    15 | pub struct Map<K, V>(MapImpl<K, V>);
       |                      ^^^^^^^^^^^^^^ the trait `core::cmp::Ord` is not implemented for `K`
    ```

- Hash and Eq are required for all four of LinkedHashMap's impls:

    ```
    error[E0277]: the trait bound `K: core::cmp::Eq` is not satisfied
      --> src/map.rs:15:22
       |
    15 | pub struct Map<K, V>(MapImpl<K, V>);
       |                      ^^^^^^^^^^^^^^ the trait `core::cmp::Eq` is not implemented for `K`
       |

    error[E0277]: the trait bound `K: core::hash::Hash` is not satisfied
      --> src/map.rs:15:22
       |
    15 | pub struct Map<K, V>(MapImpl<K, V>);
       |                      ^^^^^^^^^^^^^^ the trait `core::hash::Hash` is not implement
    ```

What I did was just implement these for \<K=String, Value=Value>. I found this to be the best balance of:

- now that Map is opaque, allowing us to tweak the internal representation for performance or other reasons; other implementations may have tighter trait bounds than BTreeMap or LinkedHashMap
- keeping the definition of Value explicit and obvious; compare:

    ```rust
    pub enum Value {
        Null,
        Bool(bool),
        Number(Number),
        String(String),
        Array(Vec<Value>),
        Object(Map<String, Value>),
    }
    ```

    vs

    ```rust
    pub enum Value {
        Null,
        Bool(bool),
        Number(Number),
        String(String),
        Array(Vec<Value>),
        Object(Map), // of...?
    }
    ```

PR #204, opened 2017-01-21
Derive Debug for Value

### LINKED ISSUE #190: Derive Debug for Value
The Debug representation is JSON, same as the Display representation. I find this surprising. Generally Debug should just be derived.

dtolnay:
Fixes #190.


dtolnay:
Nightly fails because of clippy :panda_face:.

PR #205, opened 2017-01-21
Use serde_json::Deserializer for all deserialization

### LINKED ISSUE #193: How to specialise on JSON deserialiser
Now that there's a non-public `DeserializerImpl` which is the one that's actually being used in `from_trait()`, it seems to have become impossible to use specialisation on JSON decoders when using the top-level `from_*()` methods.

E.g., I naively assumed something like this would work:

```rust
#![feature(specialization)]

struct Foo;

    fn deserialize<D>(de: &mut D) -> Result<Foo, D::Error>
        where D: serde::Deserializer
    {
        type Result<E> = Result<Foo, E>;

        trait Deserialize<D: serde::Deserializer> {
            fn deserialize(de: &mut D) -> Result<D::Error>;
        }

        impl<D: serde::Deserializer> Deserialize<D> for Foo {
            default fn deserialize(de: &mut D) -> Result<D::Error> {
                /* Default impl */
            }
        }

        impl<Iter> Deserialize<json::Deserializer<Iter>> for Foo
            where Iter: Iterator<Item=::std::io::Result<u8>>
        {
            fn deserialize(de: &mut json::Deserializer<Iter>) -> Result<json::Error> {
                /* JSON specialization */
            }
        }

        <Foo as Deserialize<D>>::deserialize(de)
    }
}
```

... and it kind of does, but the default impl is triggered when you use functions like `from_str()`, because `serde_json` doesn't use `serde_json::Deserializer` but instead a custom hidden deserialiser.

Would it make sense for `from_trait()` to somehow use `serde_json::Deserializer` instead? (apologies if I'm missing something obvious; maybe this problem is currently unavoidable)

dtolnay:
Fixes #193.

Previously the from_iter / from_reader / from_slice / from_str functions would go through the private DeserializeImpl struct which made it impossible to specialize on the JSON deserializer.

This approach makes it possible but a bit awkward because I would like to keep the `Read` trait out of the public API. The resulting code on the user's side looks something like this:

```rust
struct S<R> {
    json_deserializer: serde_json::Deserializer<R>,
}

impl<R> S<R>
    where for<'a> &'a mut serde_json::Deserializer<R>: Deserializer<Error = serde_json::Error>
{
    fn next<T: Deserialize>(&mut self) -> Result<T, serde_json::Error> {
        Deserialize::deserialize(&mut self.json_deserializer)
    }
}
```


dtolnay:
As you can tell from the tests, I also made StreamDeserializer much nicer to construct and possible to use with the &[u8] / &str optimizations that were previously only for from_slice / from_str.

```diff
  let data = "{ ... }";
- let mut stream: StreamDeserializer<Value, _> = StreamDeserializer::new(
-     data.as_bytes().iter().map(|byte| Ok(*byte))
- );
+ let mut stream = Deserializer::from_str(data).into_iter::<Value>();

dtolnay:
The nightly build fails because of clippy.

aldanor:
Thanks, great! This should make it possible to specialise in a non-hacky way

dtolnay:
I fixed the conflict.

PR #207, opened 2017-01-23
Add a tt-muncher for constructing Value

dtolnay:
This puts the `object!` and `array!` macros from json-rust and the `json!` compiler plugin from [json_macros](https://github.com/tomjakubowski/json_macros) to shame.


dtolnay:
WIP because I am porting the test suite to use this. Win-win - we get test coverage for this macro and our tests become somewhat readable.

dtolnay:
Here is the rustdoc in rendered form:

-------

Construct a `serde_json::Value` from a JSON literal.

```rust
let value = json!({
    "code": 200,
    "success": true,
    "payload": {
        "features": [
            "serde",
            "json"
        ]
    }
});
```

Any variable or expression that implements Serde's `Serialize` trait can be interpolated into the JSON literal just by referring to it.

```rust
let code = 200;
let features = vec!["serde", "json"];

let value = json!({
   "code": code,
   "success": code == 200,
   "payload": {
       features[0]: features[1]
   }
});
```

Trailing commas are allowed inside both arrays and objects.

```rust
let value = json!([
    "notice",
    "the",
    "trailing",
    "comma -->",
]);
```

bluetech:
Hi @dtolnay, this is very nice.

As a micro-optimization, do you think there is any way to use `Vec::with_capacity(3)` for something like `json!([1,2,3])` instead of `Vec::new()`?

I looked how [`vec!  does it`](https://github.com/rust-lang/rust/blob/91f34c0c70746f5c938d25d02a8a66b41240b2f0/src/libcollections/macros.rs#L45), IIUC it "cheats" by using the builtin syntax.

dtolnay:
Good idea @bluetech. I am busy until this afternoon but when I get back I will make it generate `Value::Array(vec![...])` directly so we can benefit from any optimizations it provides.

KodrAus:
Ah that's actually usable! What's the plan to release this? Are you going to cut a release or bundle it up with a bunch of things post macros 1.1?

dtolnay:
This will be in 0.9.0. I am working as hard as I can to have it out by the end of Tuesday.

dtolnay:
@bluetech thanks again, I made that change in #209. Now we will get `Value::Array(vec![...])`.

bluetech:
@dtolnay this is some admirable display of macro mastery :)

PR #210, opened 2017-01-24
Serialize T by ref

### LINKED ISSUE #151: Audit public API for missing ?Sized bounds
@withoutboats noticed that [`to_writer`](https://docs.serde.rs/serde_json/ser/fn.to_writer.html) could allow `T: ?Sized`, allowing something like `to_writer(w, "str")`.

dtolnay:
Fixes #151. Requires https://github.com/serde-rs/serde/pull/716.


oli-obk:
^^ I was almost finished with this yesterday, should have probably noted somewhere that I was working on it.

PR #211, opened 2017-01-24
Implement indexing into Value

dtolnay:
We could do IndexMut too but I looked at [what json-rust does](https://github.com/maciejhirsz/json-rust/blob/v0.11.5/src/value/mod.rs#L546-L566) and it is pretty bizarre. I think this is enough to get us most of the value of straightforward indexing.

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -470,8 +470,35 @@ mod private {
     impl<'a, T: ?Sized> Sealed for &'a T where T: Sealed {}
 }
 
+// The usual semantics of Index is to panic on invalid indexing.
+//
+// That said, the usual semantics are for things like Vec and BTreeMap which
+// have different use cases than Value. If you are working with a Vec, you know
+// that you are working with a Vec and you can get the len of the Vec and make
+// sure your indices are within bounds. The Value use cases are more
+// loosey-goosey. You got some JSON from an endpoint and you want to pull values
+// out of it. Outside of this Index impl, you already have the option of using
+// value.as_array() and working with the Vec directly, or matching on
+// Value::Array and getting the Vec directly. The Index impl means you can skip
+// that and index directly into the thing using a concise syntax. You don't have
+// to check the type, you don't have to check the len, it is all about what you
+// expect the Value to look like.
+//
+// Basically the use cases that would be well served by panicking here are
+// better served by using one of the other approaches: get and get_mut,
+// as_array, or match. The value of this impl is that it adds a way of working
+// with Value that is not well served by the existing approaches: concise and
+// careless and sometimes that is exactly what you want.
 impl<I> ops::Index<I> for Value where I: Index {
     type Output = Value;
+
+    /// Index into a `serde_json::Value` using the syntax `value[0]` or
+    /// `value["k"]`.
+    ///
+    /// Returns `Value::Null` if the type of `self` does not match the type of
+    /// the index, for example if the index is a string and `self` is an array
+    /// or a number. Also returns `Value::Null` if the given key does not exist
+    /// in the map or the given index is not within the bounds of the array.
     fn index(&self, index: I) -> &Value {
         static NULL: Value = Value::Null;
         index.index_into(self).unwrap_or(&NULL)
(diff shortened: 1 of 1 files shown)


json/src/value.rs
```diff
@@ -77,6 +78,73 @@ fn parse_index(s: &str) -> Option<usize> {
 }
 
 impl Value {
+    /// Index into a JSON array or map. A string index can be used to access a
+    /// value in a map, and a usize index can be used to access an element of an
+    /// array.
+    ///
+    /// Returns None if the type of `self` does not match the type of the index,
```
oli-obk: `None` should be in ticks

json/src/value.rs
```diff
@@ -327,6 +395,86 @@ impl Value {
     }
 }
 
+/// A type that can be used to index into a `serde_json::Value`. See the `get`
+/// and `get_mut` methods of `Value`.
+pub trait Index: private::Sealed {
```
oli-obk: is this the "idiomatic" way to make a trait unimplementable?

json/src/value.rs
```diff
@@ -327,6 +395,86 @@ impl Value {
...
+    impl Sealed for str {}
+    impl Sealed for String {}
+    impl<'a, T: ?Sized> Sealed for &'a T where T: Sealed {}
+}
+
+impl<I> ops::Index<I> for Value where I: Index {
+    type Output = Value;
+    fn index(&self, index: I) -> &Value {
+        static NULL: Value = Value::Null;
+        index.index_into(self).unwrap_or(&NULL)
```
oli-obk: The usual semantics are to panic on invalid indexing. What's your reasoning behind diverging from this pattern? Write the reasoning as a comment, I'm gonna forget again in the future ^^

json/src/value.rs
dtolnay: I got the technique from [here](https://www.reddit.com/r/rust/comments/25gvwl/prevent_another_crate_from_implementing_a_trait/) and the name from [here](https://internals.rust-lang.org/t/pre-rfc-sealed-traits/3108).

dtolnay:
I don't think visibility rules affect this at all. The Sealed trait is `pub` so it is allowed to be in the public API. That does not mean there needs to be a public path to access it.

json/src/value.rs
dtolnay: Here is what the rustdoc looks ilke:


selection_018

json/src/value.rs
dtolnay: The usual semantics are for things like Vec and BTreeMap which have different use cases than Value. If you are working with a Vec, you know that you are working with a Vec and you can get the len of the Vec and make sure your indices are within bounds. The Value use cases are more loosey-goosey. You got some JSON from an endpoint and you want to pull values out of it. Without this change, you already have the option of using `value.as_array()` and working with the Vec directly, or matching on Value::Array and getting the Vec directly. This change means you can skip that and index directly into the thing using a concise syntax. You don't have to check the type, you don't have to check the len, it is all about what you expect the Value to look like.

Basically I think the use cases that would be well served by panicking here are better served by using one of the other approaches: `get` and `get_mut`, `as_array`, or `match`. The value of this impl is that it adds a way of working with Value that is not well served by the existing approaches: concise and careless and sometimes that is exactly what you want.

I will add a comment.

dtolnay:
> **\<oli_obk_>** dtolnay: how about making it cfg(feature = "unstable")?
> **\<oli_obk_>** then we can play with it

Let's play with it now :dog:. I don't think there are so many different use cases for indexing that we can't tell whether this approach will serve us well.

json/src/value.rs
dtolnay: I added an explanation in be3d098b3c8d51400da6fee8d9761bbaadd79727.

json/src/value.rs
dtolnay: I added an explanation in 7ded5a56a1e5b92ee554f48f7cdf116f08f3725e. If we discover a different idiomatic way, we can always change it.

json/src/value.rs
dtolnay: Fixed in 77eaeeb61d4efc9429814cb7ccd692326c25f1f5.

Issue #183, opened 2017-01-03
Expose number of bytes processed by StreamDeserializer


dtolnay:
Use case from IRC:

> **\<Yorhel>** Hi everyone. Does serde_json support decoding JSON with trailing data?
> **\<Yorhel>** E.g. I'd like to do something in the form of: let (val, data) = serde_json::from_str("{}garbage");
> **\<Yorhel>** With val being the parsed Value and data pointing to the trailing "garbage"

Bincode has a bytes_read() method on their Deserializer that would support this much better than serde_json. You can kind of fake it like this:

```rust
let mut bytes = 0;
let stream = StreamDeserializer::new(s.bytes().map(|b| { bytes += 1; Ok(b) });
/* deserialize a thing */
let garbage = &s[bytes..];
```

But there must be a better way for us to support this.

dtolnay:
Specifically, I think the number we want is the number of bytes so far deserialized into a successful T.

```rust
let data = b"[0] [1] [";

let mut de = serde_json::Deserializer::from_slice(data).into_iter::<Vec<i32>>();
assert_eq!(0, de.byte_pos());

println!("{:?}", de.next()); // [0]
assert_eq!(3, de.byte_pos());

println!("{:?}", de.next()); // [1]
assert_eq!(7, de.byte_pos());

println!("{:?}", de.next()); // error
assert_eq!(8, de.byte_pos());

// if err.is_eof(), can join the remaining data to new data and continue
let remaining = &data[de.byte_pos()..];
```

dtolnay:
Here is a related use case from @vorner that I believe is addressed by byte_offset().

> **\<vorner>** Hello. I'm trying to create a JSONRPC server, so I want to combine tokio and serde_json. The problem is, I'd either need something like from_slice which would tell me how many bytes it used, or StreamDeserializer that would work on top of Futures' Stream instead of an iterator. Is there a way to get something like that?
> **\<oli_obk_>** vorner: can't you tunr a Stream into an Iterator that simply stops on encountering an error?
> **\<oli_obk_>** vorner: it might even store the error inside the iterator itself, so you can obtain it afterwards
> **\<vorner>** oli_obk_: Stops? Like ends?
> **\<vorner>** I don't worry about errors.
> **\<vorner>** The problem is, the data arrive in chunks over network.
> **\<vorner>** So it returns NotReady
> **\<oli_obk_>** ohh...
> **\<vorner>** I need to stop processing that Stream *for now* and resume later when more data arrive.
[...]

PR #212, opened 2017-01-24
allow reading from a slice that contains trailing bytes

oli-obk:
cc @vorner

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -961,12 +961,14 @@ pub fn from_partial_stream<T>(v: &[u8]) -> Result<Option<(T, usize)>>
     use super::error::ErrorCode::*;
     let mut de = Deserializer::new(read::SliceRead::new(v));
     match de::Deserialize::deserialize(&mut de) {
-        Ok(val) => Ok(Some((value, de.read.pos()))),
-        Err(EOFWhileParsingList) |
-        Err(EOFWhileParsingObject) |
-        Err(EOFWhileParsingString) |
-        Err(EOFWhileParsingValue) => Ok(None),
-        Err(other) => Err(other),
+        Ok(value) => Ok(Some((value, de.read.pos()))),
+        Err(e) => match *e.0 {
+            ErrorImpl::Syntax(EOFWhileParsingList, _, _) |
+            ErrorImpl::Syntax(EOFWhileParsingObject, _, _) |
+            ErrorImpl::Syntax(EOFWhileParsingString, _, _) |
+            ErrorImpl::Syntax(EOFWhileParsingValue, _, _) => Ok(None),
+            _ => Err(e),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


dtolnay:
Do you think we could generalize this to work with any JSON deserializer? The stream API is a lot nicer now with #205 than it used to be, and this seems like it would belong there.

Something like:

```rust
let mut stream = Deserializer::from_slice(data).into_iter::<T>();

/* deserialize some things */

let pos = stream.byte_pos();
```

vorner:
Cool, that was fast :-). I think this is what I needed. Though, I'm not sure of the method name `try_from_slice` ‒ the difference from `from_slice` is that it returns the position as well, so the try seems bit confusing.

oli-obk:
> Do you think we could generalize this to work with any JSON deserializer?

`from_iter` and `from_reader` would need to count bytes, too, but don't do it right now. So it would probably be slower then, and maybe impossible with `from_reader` entirely, since we might accidentally read part of the next value.

oli-obk:
> so the try seems bit confusing.

a failure in that function doesn't mean an actual failure, it could just be that we're missing bytes

vorner:
| a failure in that function doesn't mean an actual failure, it could just be that we're missing bytes

Thinking about this, is there any better way than to check for one of the `EOFWhileParsing` to see if it ended prematurely or if it was actually wrong?

oli-obk:
> Thinking about this, is there any better way than to check for one of the EOFWhileParsing to see if it ended prematurely or if it was actually wrong?

Oh right. I actually wanted to change that function to return a `Result<Option<(T, usize)>>`, so you'd never get an `EOFWhileParsing`, since that's impossible in an infinite stream. Basically if you get a `None` and you have no more data, that's when you're in trouble ;)

oli-obk:
> Do you think we could generalize this to work with any JSON deserializer? The stream API is a lot nicer now with #205 than it used to be, and this seems like it would belong there.

We could do that, but the ergonomics are weird, because you have to recreate the entire streamdeserializer when the previous one stops yielding elements.

The cleanest solution would probably be a specially tailored from_tokio_stream function in a tokio feature. It could internally buffer the slices until it's possible to read without triggering eof.

The "peek or null" methods make it kind of hard to distinguish between errors and end of stream, so these would have to be touched. Not sure how much they affect performance.

dtolnay:
@vorner for your use case, if the input is `b"1" as &[u8]` would you expect it to return `Ok(None)` or `Ok(Some((Number(1), 1)))`? There isn't an immediate way to tell whether this will be followed by more digits in the same number.

dtolnay:
I think we need this function to reject types other than array and object.

vorner:
Hmm, number is stupid :-|. It can work with true/false/nil and with strings as well as arrays and objects. But I don't know about numbers.

Actually, all things that usually require this have objects or arrays, but I'm not sure what to do with this corner case. Maybe just return `Some(1)` and let the caller decide what to do (with some kind of warning in the documentation)?

dtolnay:
Closing in favor of #183.

PR #218, opened 2017-01-26
Index into Map

### LINKED ISSUE #217: Implementing Index on serde_json::Map
Indexing on Value is very nice ! However, it'd be even nicer if serde_json::Map could be indexed in a similar way. I have some code where I match over a Value that looks like this :

```rust
match *val {
    Value::String(ref s) => s,
    Value::Array(ref arr) => arr[0].as_str().unwrap(),
    Value::Object(_) => val["type"].as_str().unwrap(),
    _ => panic!()
}
```

It'd be a bit nicer if I could instead write

```rust
match *val {
    Value::String(ref s) => s,
    Value::Array(ref arr) => arr[0].as_str().unwrap(),
    Value::Object(ref map) => map["type"].as_str().unwrap(),
    _ => panic!()
}
```

dtolnay:
Fixes #217.

### DIFF
--- a/json/src/map.rs
+++ b/json/src/map.rs
@@ -3,6 +3,7 @@ use std::fmt::{self, Debug};
 use value::Value;
 use std::hash::Hash;
 use std::borrow::Borrow;
+use std::ops;
 
 #[cfg(not(feature = "preserve_order"))]
 use std::collections::{BTreeMap, btree_map};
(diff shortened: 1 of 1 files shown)


dtolnay:
Ping @oli-obk.

I think this `String: Borrow<Q>, Q: Ord + Eq + Hash` bound may have been a mistake because it limits our ability to swap in other map implementations transparently. Maybe we should have used a simpler bound like `AsRef<str>`. But we already used this bound for `get` and `get_mut` and `contains_key` so it is better to be consistent.

oli-obk:
> may have been a mistake

we can revisit this for the next breaking release. It's a subtle change, that is unlikely to be noticed by anyone.

PR #229, opened 2017-01-29
Make tests more readable with raw strings

killercup:
Or maybe I'm just allergic to `\"`s?

### DIFF
--- a/json_tests/tests/test.rs
+++ b/json_tests/tests/test.rs
@@ -171,8 +171,8 @@ fn test_encode_nonfinite_float_yields_null() {
 #[test]
 fn test_write_str() {
     let tests = &[
-        ("", "\"\""),
-        ("foo", "\"foo\""),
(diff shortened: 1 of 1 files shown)


dtolnay:
Thanks! In 40a3406b5e959ceeba31a090d204cb16fddb845b I added a `json_str!` macro to handle this in an even more readable way. You write JSON and it turns it into a sequence of `concat!` calls to give you a `&'static str` of JSON:

```diff
-            "[false,null,[\"foo\\nbar\",3.5]]",
+            json_str!([
+                false,
+                null,
+                [
+                    "foo\nbar",
+                    3.5
+                ]
+            ]),
```

I haven't gotten a chance yet to update the existing `\"` code but I would prefer to use `json_str!(...)` rather than `r#"..."#` because I can get brace matching and indentation from my IDE.

PR #230, opened 2017-01-29
Fix #220 (invalid enum parser)

### LINKED ISSUE #220: Enums can parse invalid JSON
```rust
#[macro_use]
extern crate serde_derive;
extern crate serde_json;

#[derive(Debug, Deserialize)]
enum E {
    V(u8),
}

fn main() {
    println!("{:?}", serde_json::from_str::<E>(r#" "V"0 "#));
}
```

```
Ok(V(0))
```

killercup:


json_tests/tests/test.rs
```diff
@@ -1133,6 +1133,8 @@ fn test_parse_enum_errors() {
             "trailing characters at line 1 column 9"),
         ("\"Frog\"",
             "EOF while parsing a value at line 1 column 6"),
+        ("\"Frog\" 0 ",
+            "EOF while parsing a value at line 1 column 6"),
```
killercup: I'm not sure about what `ErrorCode` to use. `EOFWhileParsingValue` seemed somewhat appropriate as it keeps the error for message for the `"\"Frog\""` case, but looking at this test, I could also add a new variant. Any preferences?

dtolnay: Hmm it is not clear to me why the `"\"Frog\""` case is an EOF error. It seems like that should be an invalid type error because the data contains a string and the enum expected a tuple variant. I will follow up on that.

PR #231, opened 2017-01-29
Serialize Map with integer keys

### LINKED ISSUE #221: Integer keys in to_value
The `to_string` function supports integer map keys but `to_value` does not.

killercup:
by transforming them to strings.

Fixes #221


killercup:
Okay, @dtolnay, that's the last one for now. Sorry for opening a truckload of PRs for relatively minor changes; they seemed separate enough to not lump them together in one big PR. This way, you can pick and choose what you like :)

Issue #219, opened 2017-01-27
Implement PartialEq for Value to compare with native data types Value represents


imp:
For example for the following to work
```rust
        let n = Value::from(42);
        assert_eq!(n, 42);

        let s = Value::String(String::from("hello"));
        assert_eq!(s, "hello");
        assert_eq!(s, String::from("hello"));
```

KodrAus:
@oli-obk I guess that would be `false`? What I'm wondering is what would be the result of `Value::from(42), "42"`?

oli-obk:
@KodrAus that's "obviously" false, too.

imp:
The idea was to allow compare with the native types that are being representing by Value. So that
```rust
let n = Value::from(42);
assert_ne!(n, "42");
assert_eq!(n, 42);
```

All the other fancy possible conversion are out of scope.

dtolnay:
I like this idea and I think @imp's suggested semantics make sense.

imp:
Closed via #235

PR #235, opened 2017-01-30
Implement PartialEq for Value

imp:
Addresses #219

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -427,7 +427,7 @@ impl Value {
     ///
     ///     // "Steal" ownership of a value. Can replace with any valid Value.
     ///     let old_x = value.pointer_mut("/x").map(|x| mem::replace(x, Value::Null)).unwrap();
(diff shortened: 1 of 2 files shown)


json/src/value.rs
```diff
@@ -625,6 +625,137 @@ impl<I> ops::Index<I> for Value where I: Index {
...
+}
+
+
+impl PartialEq<Value> for String {
+    fn eq(&self, other: &Value) -> bool {
+        other.as_str().map_or(false, |s| s == self)
+    }
+}
+
+macro_rules! partialeq_ixx {
```
oli-obk: you can merge all these macros, by passing the `as_*` function name and the `as *` cast type as macro arguments.

imp: Right, good advice!

imp:
@oli-obk Thanks for review. Indeed much cleaner now.

oli-obk:
thanks

PR #243, opened 2017-02-04
Generalize json! macro to ToJson

### LINKED ISSUE #242: to_value should be bound on ToJson, not Serialize
All types that implement `Serialize` implement `ToJson`, but not all types that implement `ToJson` necessarily implement `Serialize`. So `to_value()` should be bound on `ToJson`, not `Serialize`, since that's all it really needs. That way I can implement `ToJson` on my custom types and then use them with the `json!()` macro.

For reference, I'm implementing `ToJson` on my type and not `Serialize` because my type can actually be serialized to both JSON and XML, but the serializations have very different representations, and I don't want to confuse anyone.

dtolnay:
Fixes #242. cc @kballard

### DIFF
--- a/json/src/macros.rs
+++ b/json/src/macros.rs
@@ -254,6 +254,6 @@ macro_rules! json_internal {
     // Any Serialize type: numbers, strings, struct literals, variables etc.
     // Must be below every other rule.
     ($other:expr) => {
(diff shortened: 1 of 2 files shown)


dtolnay:
ping @oli-obk

PR #252, opened 2017-02-11
Arbitrary precision numbers

### LINKED ISSUE #18: Parser cannot read arbitrary precision numbers
http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf specifies in the second paragraph of the introduction that JSON is agnostic about numbers, and simply represents them as a series of digits.

However, serde-json parses anything with a decimal point as a Rust `f64`, which causes numbers to be read incorrectly. There is no way to avoid this because this behaviour is chosen as soon as a decimal point is encountered. This makes it impossible to use serde-json to interoperate with financial software using JSON.

dtolnay:
Fixes #18.

```toml
serde_json = { version = "0.9", features = ["arbitrary_precision"] }
```

```rust
#[derive(Serialize, Deserialize)]
struct S {
    n: serde_json::Number,
    v: serde_json::Value,
}

let s: S = serde_json::from_str(...)?;

// full precision
println!("{}", s.n);
println!("{}", s.v);
println!("{}", serde_json::to_string(&s)?);
```


dtolnay:
@alexreg please take a quick look at this and let us know whether it satisfies your use case.

alexreg:
A few name changes I would suggest for clarity/consistency:

`ParseNumber` -> `Json(Number)Visitor`
`VisitNumber` -> `(Number)Visitor`
`SerializeNumber` -> `NumberSerializer`

Take them or leave them. :)

dtolnay:
Thanks for the idea. I would prefer to leave it as is. We use `...Visitor` for things that implement the Visitor trait and `...Serializer` for things that implement the Serializer trait, which is not the case here.

dtolnay:
Waiting on https://github.com/serde-rs/json/issues/18#issuecomment-279270898.

alexreg:
Fair enough.

dtolnay:
I am going to try to rework this to use the same trick that TOML uses for specialized handling of Datetimes without relying on specialization.

alexreg:
If it's using a "trick", I'd be very much against that! Nothing too hacky, please.

dtolnay:
Let's call it "under-explored abilities of the API" rather than "trick."

alexreg:
Okay. I quite like the way it worked currently... so I hope it doesn't change too much!

alexreg:
Any update on this yet, @dtolnay?

dtolnay:
Nope, will be post 1.0.

alexreg:
Okay. No chance of getting a feature-gated specialisation-based version before then? I don't know when 1.0 will be, but I guess soon-ish...

dtolnay:
No chance of getting a feature-gated specialization-based version. We are making some lifetime-related changes in the next release (#288) and specialization interacts very poorly with lifetimes - use after free, corrupted data, etc. See https://github.com/rust-lang/rust/issues/40582.

alexreg:
@dtolnay How about now? :)

PR #253, opened 2017-02-13
Implement a bunch of `From`s

### LINKED ISSUE #241: Implement From on Value for more types
`Value` implements `From` for numeric types. It should also implement `From` for any type that can be represented as a JSON value, such as `Map<String, Value>`, `bool`, `Vec<Value>` (and probably `&'a [Value]` as well), and in particular, for `String`, `&'a str`, and `Cow<'a, str>`. That way I can write code like

```rust
map.insert("key".into(), someStr.into())
```

instead of

```rust
map.insert("key".into(), serde_json::Value::String(someStr.into()))
```

killercup:
This is how I spend my lunch breaks! 😆 (Mostly copy-pasta, though.)

I'd love to `impl<T: Into<String>> From<T> for Value` and probably also `impl<T: Iterator<Item=I>, I: Into<Value> From<T> for Value`. Maybe with specialization?

Fixes #241

### DIFF
--- a/json/src/value.rs
+++ b/json/src/value.rs
@@ -87,6 +87,7 @@ use std::io;
 use std::ops;
 use std::str;
 use std::vec;
+use std::borrow::Cow;
 
 use serde::de::{self, Unexpected};
 use serde::ser;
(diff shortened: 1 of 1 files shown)


json/src/value.rs
```diff
@@ -812,17 +813,181 @@ from_integer! {
 }
 
 impl From<f32> for Value {
+    /// Convert 32-bit floating point number to `Value`
```
oli-obk: do documented impls show up in rustdoc at all?

killercup: Yes, likes this:

bildschirmfoto 2017-02-13 um 15 03 16

(Also, the doc tests are nicer than unit tests here IMHO.)

oli-obk:
> This is how I spend my lunch breaks!

/me hides shamefully for being lazy

json/src/value.rs
```diff
@@ -812,17 +813,181 @@ from_integer! {
...
+    /// Convert string slice to `Value`
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # extern crate serde_json;
+    /// use serde_json::Value;
+    ///
+    /// # fn main() {
+    /// let s: String = "lorem".to_string();
```
oli-obk: too much pasta

killercup: Oops. I meant "ipsum", obviously :trollface:

killercup: Fixed. Kept the "lorem", though.

oli-obk:
Do you think it would be totally crazy to implement [`FromIterator<T>`](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) for `Value`, yielding `Value::Array` for `T: Into<Value>`?

Then we could do `someiterator.collect()` and get a `Value::Array`. Seems cleaner to me than that `Clone` bound.

killercup:
@oli-obk, "crazy", you say? "DONE!", I answer! I did not remove the Vec and slice impl, though, as they are still a bit more ergonomic and you don't need to `use std::iter::FromIterator;`.

Feel free to pick and choose what you'd like to have :)

PR #255, opened 2017-02-16
Minor doc formatting improvements for `Number`.

frewsxcv:


dtolnay:
I agree with most of this except for `` `Number` ``. For example in is_i64, the point is not to represent the serde_json::Number as i64. The point is to represent the [JSON number](https://tools.ietf.org/html/rfc7159#section-6) (or [number](https://en.wikipedia.org/wiki/Number), depending on your perspective) as i64. That is, serde_json::Number and i64 are both just representations of some underlying concept of a number. In is_i64 or as_i64, we are taking that underlying concept and turning it into a different representation. Your emphasis makes it seem like the i64 represents a serde_json::Number which represents a JSON number, which does not seem accurate.

frewsxcv:
Addressed in latest force push.

json/src/number.rs
```diff
@@ -82,7 +82,7 @@ impl Number {
         }
     }
 
-    /// Converts a finite f64 to a Number. Infinite or NaN values are not JSON
+    /// Converts a finite `f64` to a number. Infinite or NaN values are not JSON
```
dtolnay: Oops you had this one right before. This one really does refer to a `serde_json::Number`.

frewsxcv:
Addressed in latest force push.

dtolnay:
Thanks!

PR #257, opened 2017-02-19
Fix #256: support deserializing strings with arbitrary bytes

### LINKED ISSUE #256: Support deserializing arbitrary bytes inside strings
I know that it's not correct according to the JSON standard, but sometimes you get to deal with json that has arbitrary bytes inside a string literal (except '"'). To accomodate this use case, I suggest that we explictly implement the following two functions for the Deserializer:

* deserialize_bytes
* deserialize_byte_buf

to parse Strings without doing any processing on them, just returning the bytes. They can then be used to explictly request that one wants to parse a "raw" string.

bennofs:
Usecase described in #256. Also see the included documentation.

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -649,9 +649,72 @@ impl<'a, R: Read> de::Deserializer for &'a mut Deserializer<R> {
         }
     }
 
+    /// Parses a JSON string as bytes. Note that this function does not
+    /// check whether the bytes represent valid unicode code points.
+    ///
+    /// The JSON specification requires that strings only contain valid
+    /// unicode characters. To deal with non-conforming JSON, you may use
+    /// this function, which attempts to parse a string without checking
+    /// whether the bytes represent valid unicode code points.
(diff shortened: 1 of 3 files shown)


dtolnay:
I misinterpreted "without doing any processing" in your proposal - I was thinking processing as in performing UTF-8 validation. Without interpreting escape sequences, this is a breaking change because the string containing backslash+n used to deserialize as one byte and now it deserializes as two bytes. Would this still be useful to you if we interpret escape sequences but just don't do the final from_utf8 conversion?

bennofs:
OK, that works for my usecase as well. I thought that it might be a good idea to go all the way and allow the user to request access to the low level bytes, but I was not aware that other functionality uses visit_bytes already.

bennofs:
@dtolnay changed it to only skip unicode validation and to allow delegate to `deserialize`.

dtolnay:
Thanks, I released this in serde_json 0.9.7.

Issue #259, opened 2017-02-20
Impossible to store a struct field of type serde_json::StreamDeserializer<?, T>


dflemstr:
While working on https://github.com/dflemstr/rq/issues/151 I noticed that `serde_json::Deserializer::from_reader(r).into_iter()` returns a Voldemort type, so it can't be stored in a struct.  I want to create a:

```rust
pub struct JsonSource<R>(serde_json::StreamDeserializer<R, value::Value>)
    where R: serde_json::de::Read;
```

...but `R` is not actually a variable, but rather the private type `IteratorRead<Bytes<R2>>` for a type `R2: std::io::Read`.  The `IteratorRead` struct would have to be public for the type to be nameable.

dtolnay:
I just fixed this earlier today in serde_json 0.9.7. Your code works for me:

```rust
extern crate serde_json;
use serde_json::value;

pub struct JsonSource<R>(serde_json::StreamDeserializer<R, value::Value>)
    where R: serde_json::de::Read;

fn main() {}
```

dflemstr:
Try instantiating the struct with a `serde_json::Deserializer::from_reader(r).into_iter()` in it and you will see what I mean!

dtolnay:
Can you share your code? I'm still not getting it.

```rust
pub fn f() {
    let read = File::open("/etc/passwd").unwrap();
    let _ = JsonSource(serde_json::Deserializer::from_reader(read).into_iter());
}
```

dflemstr:
Sorry for not creating a more complete example at first, I thought the description would be enough.  Basically, in your example you avoid naming the type since you don't specify the type of the `_`.  But this is my situation:

```rust
extern crate serde_json;

use std::io;

use serde_json::value;

pub struct JsonSource<R>(serde_json::StreamDeserializer<R, value::Value>)
    where R: serde_json::de::Read;

fn source<IoRead>(r: R) -> JsonSource<???>
    where IoRead: io::Read
{
    serde_json::Deserializer::from_reader(r).into_iter()
}
```

What is `???`?

dtolnay:
You're right this is still bad. I will expose all the types.

dtolnay:
Addressed in #260. I'll cut a release tomorrow morning.

dtolnay:
Released in serde_json 0.9.8.

PR #260, opened 2017-02-21
Expose the various Read implementations

dtolnay:
@dflemstr can you confirm that this would fix your use case in #259?

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -10,7 +10,7 @@ use super::error::{Error, ErrorCode, Result};
 
 use read;
 
-pub use read::Read;
+pub use read::{Read, IteratorRead, SliceRead, StrRead};
 
 //////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


json/src/de.rs
```diff
@@ -21,8 +21,19 @@ pub struct Deserializer<R> {
...
+    /// Create a JSON deserializer from one of the possible serde_json input
+    /// sources.
+    ///
+    /// Typically it is more convenient to use one of these methods instead:
+    ///
+    ///   - Deserializer::from_str
+    ///   - Deserializer::from_bytes
+    ///   - Deserializer::from_iter
+    ///   - Deserializer::from_reader
+    pub fn new(read: R) -> Self {
```
dflemstr: Please note that for my use case, exposing just the types would be enough and these methods could still be private.

oli-obk:
Exposing these functions has the advantage of confusing fewer ppl who upgrade.

PR #261, opened 2017-02-21
Deserializer for &Value

### LINKED ISSUE #195: Implement Deserializer for &Value
We have an implementation of Deserializer for Value which consumes the Value - for example Value::String turns into visit_string. It would make sense to also implement Deserializer for &Value, so that for example &Value::String turns into visit_str.

dtolnay:
Fixes #195.

### DIFF
--- a/json/src/number.rs
+++ b/json/src/number.rs
@@ -180,6 +180,27 @@ impl Deserializer for Number {
     }
 }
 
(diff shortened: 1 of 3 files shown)


oli-obk:
Needs a test. Otherwise lgtm

PR #265, opened 2017-02-21
Fix regressions introduced by #256 fix

bennofs:
I'm sorry, my fix for #256 had two errors:

* I forgot to clear the str_buf before parsing bytes. This results in incorrect parses.
* Somehow, my patch regressed performance by a lot. I'm not sure why this is the case, but using `peek_or_null` instead of `peek` fixes the issue.

The release with the previous PR should probably be yanked at crates.io. I apologize for the trouble, I should have verified my code better.

I also added a test for this particular failure case.

### DIFF
--- a/json/src/de.rs
+++ b/json/src/de.rs
@@ -706,6 +706,7 @@ impl<'a, R: Read> de::Deserializer for &'a mut Deserializer<R> {
         match try!(self.peek_or_null()) {
             b'"' => {
                 self.eat_char();
+                self.str_buf.clear();
                 let slice = try!(self.read.parse_str_raw(&mut self.str_buf));
                 visitor.visit_bytes(slice)
             }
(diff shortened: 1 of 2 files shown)


bennofs:
Oh, I found why the performance regression occurred: The new code always called `self.peek_error` to construct an error for the call to `ok_or`. `peek_error` calls `peek_position` which is an expensive operation.

json/src/de.rs
```diff
@@ -699,14 +699,14 @@ impl<'a, R: Read> de::Deserializer for &'a mut Deserializer<R> {
     fn deserialize_bytes<V>(self, visitor: V) -> Result<V::Value>
         where V: de::Visitor
     {
-        try!(self.parse_whitespace());
-
-        let next = try!(self.peek());
-        let next = try!(next.ok_or(self.peek_error(ErrorCode::ExpectedSomeString)));
+        if try!(self.parse_whitespace()) { // true if eof
+            return Err(self.peek_error(ErrorCode::EOFWhileParsingValue));
```
dtolnay: We don't need this check right? deserialize_enum and deserialize_option don't have it. If we hit EOF, the peek_or_null will return '\0' so it goes to self.deserialize(visitor) which handles this error.

bennofs: yes, that is correct. I added it only to ensure the code is as similar as possible to `parse_value`, so correctness is easy to see. But it can be removed of course.

PR #267, opened 2017-02-27
Redefine Serializer::collect_str

nox:


nox:
Related to https://github.com/serde-rs/serde/pull/789, do not merge yet.

json/src/ser.rs
```diff
@@ -307,6 +308,50 @@ impl<'a, W, F> ser::Serializer for &'a mut Serializer<W, F>
...
+            error: Option<Error>,
+        }
+
+        impl<'ser, W, F> Write for Adapter<'ser, W, F>
+            where W: io::Write,
+                  F: Formatter,
+        {
+            fn write_str(&mut self, s: &str) -> fmt::Result {
+                assert!(self.error.is_none());
+                match self.formatter.write_string_fragment(&mut self.writer, s.as_bytes()) {
```
dtolnay: The `write_string_fragment` method dumps bytes directly into the output. Instead you need to factor out the middle part of `format_escaped_str` and use that here to properly escape any special characters.

nox: Made `write_string_fragment` take a `&str` and contain the whole loop from the function `format_escaped_str`.

nox:
I also added a commit to optimise serialising a single `char.

dtolnay:
This is a breaking change for rq which overrides write_string_fragment: [rq/value/json.rs](https://github.com/dflemstr/rq/blob/v0.9.2/src/value/json.rs)

Please do not make any changes to Formatter in this PR. The loop for calling write_string_fragment / write_char_escape should live outside of Formatter.

I filed #268 to follow up on the argument type in the next major release.

json/src/ser.rs
```diff
@@ -1245,11 +1288,11 @@ fn format_escaped_char<W: ?Sized, F: ?Sized>(wr: &mut W, formatter: &mut F, valu
...
-    // FIXME: this allocation is required in order to be compatible with stable
-    // rust, which doesn't support encoding a `char` into a stack buffer.
-    let mut s = String::new();
-    s.push(value);
-    format_escaped_str(wr, formatter, &s)
+    use std::io::Write;
+    let mut buf = [0; 4];
+    write!(&mut buf[..], "{}", value).unwrap();
+    let str = unsafe { str::from_utf8_unchecked(&buf[0..value.len_utf8()]) };
+    format_escaped_str(wr, formatter, str)
```
dtolnay: Please do not call any variables `str`. GitHub shows it as a keyword and makes it look like a bug.

json/src/ser.rs
```diff
@@ -1245,11 +1288,11 @@ fn format_escaped_char<W: ?Sized, F: ?Sized>(wr: &mut W, formatter: &mut F, valu
...
 {
-    // FIXME: this allocation is required in order to be compatible with stable
-    // rust, which doesn't support encoding a `char` into a stack buffer.
-    let mut s = String::new();
-    s.push(value);
-    format_escaped_str(wr, formatter, &s)
+    use std::io::Write;
+    let mut buf = [0; 4];
+    write!(&mut buf[..], "{}", value).unwrap();
+    let str = unsafe { str::from_utf8_unchecked(&buf[0..value.len_utf8()]) };
```
dtolnay: Add a brief justification of why this is safe along with a link to <a href="https://github.com/serde-rs/json/issues/270"> https<span></span>://github.com/serde-rs/json/issues/270</a>.

nox: Added that.

json/src/ser.rs
nox: Renamed to `slice`.

nox:
Note that if the `Formatter` methods produced `io::Error` errors, or at least if the portion that handles strings did, we wouldn't need to implement `fmt::Write` and store a `serde_json::Error` out of band, instead we could use `write!` directly and use the returned `io::Error`.

PR #269, opened 2017-02-28
Pass &str to write_string_fragment

### LINKED ISSUE #268: write_string_fragment should take &str instead of &[u8]
https://docs.serde.rs/serde_json/ser/trait.Formatter.html#method.write_string_fragment

rq is using unsafe to get a &str: https://github.com/dflemstr/rq/blob/v0.9.2/src/value/json.rs#L224

```rust
let s = unsafe { str::from_utf8_unchecked(fragment) };
```

It should be our responsibility to guarantee that the fragment is UTF-8.

cc @dflemstr

dtolnay:
Fixes #268.

### DIFF
--- a/json/src/ser.rs
+++ b/json/src/ser.rs
@@ -910,10 +910,10 @@ pub trait Formatter {
     /// Writes a string fragment that doesn't need any escaping to the
     /// specified writer.
     #[inline]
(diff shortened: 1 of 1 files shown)


dtolnay:
Let's revisit this for the next major release.

PR #272, opened 2017-03-01
Update error.rs

tredoe:


dtolnay:
Debug is already implemented [here](https://github.com/serde-rs/json/blob/v0.9.8/json/src/error.rs#L216), which is why the CI is failing.

Can you explain the use case for PartialEq in more detail? Maybe show example code that you would like to be able to write.

tredoe:
It's to testing errors; to be able sure than a function returns a specific error.

http://play.integer32.com/?gist=da92f0b0c78221a1c2eb40f826560161&version=stable

dtolnay:
Unfortunately io::Error is not equality comparable - see [this example](https://play.rust-lang.org/?gist=677ad175edd0bcccb01871f1903dea22).

As we do in our unit tests, would it be possible for you to compare the strings instead?

```rust
assert_eq!(get_value(data).unwrap_err().to_string(), "invalid Json format");
```

Issue #279, opened 2017-03-12
Value pretty print when formatting using alternative form `{:#}`


imp:
It may be nice to control how the Value is printed using alternative formatting `{:#}`
Currently Value is printed using compact style, no matter whether normal (`{}`) or alternative form is used. It may be clever to emit pretty string when alternative form (`{:#}`) is requested in `format!` or `print!` macro.

dtolnay:
Fixed in #280.

PR #280, opened 2017-03-12
Pretty print Value when alternate formatting requested

imp:
Possible implementation of #279


imp:
Travis build needs to be respun once nightly cargo is fixed

Issue #173, opened 2016-12-08
Path to errors


dtolnay:
Newer versions of Jackson are really helpful about giving you the full path to the thing that errored out. See if this is something we can do with minimal performance impact or behind a flag.

dtolnay:
This repo includes a relatively cheap way to track the path within a Deserializer:

https://github.com/dtolnay/serde-ignored

Let's factor the path-tracking part out and use it for errors in serde_json.

dtolnay:
I implemented this for serde_yaml in https://github.com/dtolnay/serde-yaml/pull/43. Just need to port it over to JSON.

allistercsmith:
Hey @dtolnay 👋 Have you managed to make any headway on this change? I see you mentioned in https://github.com/serde-rs/serde/pull/710 that you had a design in mind, but that was quite a while ago - can't see any other updates since then.

I'm currently building an API in Rust, which will be publicly consumable once launched. Having clear error messages (i.e., path + error) is quite a fundamental piece of this.

I'm quite new to the Rust scene, but I'd be happy to try get my hands dirty if there's nothing in progress here? I was also looking at https://github.com/dtolnay/serde-ignored for inspiration.

dtolnay:
Here is the library: https://github.com/dtolnay/path-to-error. You can use it to wrap any Deserializer and find out where an error occurs.

matklad:
This is a really useful library, just helped me to instantly pin-point a bug right now, thanks!

Would be lovely to have this built-in into `serde-json`, both to improve visibility, and to reduce the amount of patching required if deserialziation happens in the dependency :)

ansrivas:
> This is a really useful library, just helped me to instantly pin-point a bug right now, thanks!
> 
> Would be lovely to have this built-in into `serde-json`, both to improve visibility, and to reduce the amount of patching required if deserialziation happens in the dependency :)

Thank you for this library :)

This just happened in one of the projects - if I am the client invoking other APIs, I quickly introduced `path-to-error` and now I can very easily debug errors.
But I am using other projects which simply use `serde-json` for e.g. actix-web uses it to deserialize incoming json requests. Would be really great if this gets included in the base lib ( behind a feature flag if needed ).

cole-abbeduto-particle:
So any plans to make this standard and to update the default Display impl to include the full path when missing fields?

PR #281, opened 2017-03-13
Conversion to io::Error

dtolnay:
@kixunil @nox any thoughts?

```rust
use std::io;

enum MyError {
    Io(io::Error),
    Json(serde_json::Error),
}

impl From<serde_json::Error> for MyError {
    fn from(err: serde_json::Error) -> MyError {
        use serde_json::error::Category;
        match err.classify() {
            Category::Io => {
                MyError::Io(err.into())
            }
            Category::Syntax | Category::Data | Category::Eof => {
                MyError::Json(err)
            }
        }
    }
}
```

### DIFF
--- a/json/src/error.rs
+++ b/json/src/error.rs
@@ -126,6 +126,51 @@ pub enum Category {
     Eof,
 }
 
+impl From<Error> for io::Error {
+    /// Convert a `serde_json::Error` into an `io::Error`.
+    ///
+    /// JSON syntax and data errors are turned into `InvalidData` IO errors.
+    /// EOF errors are turned into `UnexpectedEof` IO errors.
+    ///
+    /// ```rust
+    /// use std::io;
+    ///
+    /// enum MyError {
+    ///     Io(io::Error),
+    ///     Json(serde_json::Error),
+    /// }
+    ///
+    /// impl From<serde_json::Error> for MyError {
+    ///     fn from(err: serde_json::Error) -> MyError {
+    ///         use serde_json::error::Category;
+    ///         match err.classify() {
+    ///             Category::Io => {
+    ///                 MyError::Io(err.into())
+    ///             }
+    ///             Category::Syntax | Category::Data | Category::Eof => {
+    ///                 MyError::Json(err)
+    ///             }
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    fn from(j: Error) -> Self {
+        if let ErrorCode::Io(err) = j.err.code {
+            err
+        } else {
+            match j.classify() {
+                Category::Io => unreachable!(),
+                Category::Syntax | Category::Data => {
+                    io::Error::new(io::ErrorKind::InvalidData, j)
+                }
+                Category::Eof => {
+                    io::Error::new(io::ErrorKind::UnexpectedEof, j)
+                }
+            }
+        }
+    }
+}
+
 #[derive(Debug)]
 struct ErrorImpl {
     code: ErrorCode,
--- a/json_tests/tests/test.rs
+++ b/json_tests/tests/test.rs
@@ -18,11 +18,12 @@ use std::collections::BTreeMap;
 use std::{f32, f64};
 use std::fmt::{self, Debug};
 use std::{i8, i16, i32, i64};
+use std::io;
 use std::iter;
 use std::marker::PhantomData;
 use std::{u8, u16, u32, u64};
 
-use serde::de;
+use serde::de::{self, Deserialize};
 use serde::ser::{self, Serialize, Serializer};
 use serde::bytes::{ByteBuf, Bytes};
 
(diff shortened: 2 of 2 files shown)


json/src/error.rs
```diff
@@ -126,6 +126,51 @@ pub enum Category {
...
+    ///             Category::Syntax | Category::Data | Category::Eof => {
+    ///                 MyError::Json(err)
+    ///             }
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    fn from(j: Error) -> Self {
+        if let ErrorCode::Io(err) = j.err.code {
+            err
```
oli-obk: You could match over j.err.code and j.classify () simultaneously, to get rid of the unreachable!

dtolnay:
```rust
match (j.err.code, j.classify()) {
    (ErrorCode::Io(err), _) => err,
    (_, Category::Syntax) | (_, Category::Data) => {
        io::Error::new(io::ErrorKind::InvalidData, j)
    }
    (_, Category::Eof) => {
        io::Error::new(io::ErrorKind::UnexpectedEof, j)
    }
}
```

Where do you handle `(ErrorCode::NotIo, Category::Io)` if not unreachable?

oli-obk: oooh... dunnow what I was thinking.

Kixunil:
I think this makes the code unnecessarily complicated and doesn't represent the logic in types. Why do you want to implement it this way instead of what I suggested? Or why not just make `ErrorCode` public and provide conversion?

dtolnay:
@Kixunil see [this forum thread](https://users.rust-lang.org/t/error-types-as-enums-or-as-structs/9797/4?u=dtolnay).

We used to have a fully public error type and it was a headache, for example [here](https://github.com/serde-rs/json/pull/163/files#r85468757). It is not something that we could support in a 1.0 version of this library while actively developing it.

Another example of how public error types go wrong is it would make it impossible to implement https://github.com/serde-rs/json/issues/173. In that ticket I intend to expose path to errors, so if you have JSON that looks like `{"a":{"b":0}}` and structs `Kixunil { a: A }` and `A { b: String }`, you can call `err.path()` or something and it will tell you the error occurred while deserializing `a.b`.

I am open to other approaches that are less complicated and represent the logic in types but making everything public is not viable. Let me know if you have other ideas.

One thing that would be helpful is if you could describe exactly why you need this functionality and how you intend to use it. That way both of us can be optimizing for the same use case.

Kixunil:
@dtolnay OK, that makes sense. If I understand it correctly, the point is to allow future extensions. That seems like good idea to me. On the other hand, just providing error message is insufficient at least for cases when the message should be localised or serialised into format that preserves the structure of the error (e.g. for binary logs).

What I think might be good idea is to provide conversion from "opaque" error to "simple" error. "Simple" error should be something that is well-structured and guaranteed to always be supported. The `Category` is the example of this "simple" error. What I think is that it's current implementation is "too simple" error.

It's "too simple" because it doesn't carry information about the type, so one might accidentally write:

```rust
match err.classify() {
    Category::Syntax => handle_io_error(err.into()),
    //...
}
```

One thing that I love about Rust is it's ability to prevent these kinds of mistakes by static analysis. Interestingly, many things that we have in Rust are possible in C++ too. For example mandatory null checks is possible in C++ and conceptually similar to `Option<&T>`. So why nobody uses it? Because of C++ culture. If C++ programmers used appropriate techniques from the beginning it would be whole different world now. Rust is very young language and we are shaping it's future now. I think we should make it as good as we can.

I think that we can safely assume that any deserialisation can fail due to underlying IO error. This holds true for any deserialiser of anything out there. The only way IO error would not be present is reading from memory. This however is property of data source, not property of parser. It can be represented as generic where `!` can be used as error type for data in memory.

So for any serialisation format, deserialisation error should be convertible to:

```rust
enum SimpleError<IOE> {
    Io(IOE),
    // other possible variants
}
```

In case of Json, we can be sure that there might be data error. So the enum should look like this:

```rust
enum SimpleError<IOE> {
    Io(IOE),
    Data(DataError),
}
```

This error representation will always be appropriate, no matter what.

`DataError` can be opaque again with possibility to convert it to something simple once we decide which errors may always happen. Or we can decide right away.

`DataError` might be `SyntaxError` (syntax of Json itself was invalid) or `TypeError` (e.g. int instead of string)

`SyntaxError` could be `UnexpectedEnd`, `UnexpectedToken`

I'm not proposing to create a complicated nested type. I'm fine with having flat enum. What I want to point out is that there is probably some minimal type that will never have to change. Conversion to this minimal type should be available in the library. If new information is added later, the conversion will simply discard it to remain backward-compatible.

It might be desirable to have C-like enum (without data) too. That might be achieved by separate enum or descriptor.

Regarding generic IO error, I'd love it to be there but I can understand if there is just `io::Error`.

PR #295, opened 2017-04-11
Remove ToJson trait

### LINKED ISSUE #294: Remove ToJson trait
This trait was ripped directly from [`rustc_serialize::json::ToJson`](https://docs.rs/rustc-serialize/0.3.23/rustc_serialize/json/trait.ToJson.html) but the point was lost by providing an impl ToJson for T where T: Serialize. The rustc_serialize trait is infallible and there are impls for standard library types that can be infallibly converted to Json - not for all types that implement Encodable.

I don't think the trait as it currently exists serves a purpose. The fallible use case is covered by serde_json::to_value and the literal use case is covered by the `json!` macro.

I would be open to seeing an infallible ToJson trait explored in its own crate - ideally complete with its own custom derive which is distinct from serde's Serialize.

```rust
#[derive(ToJson)]
struct S {
    a: u8,
    b: u8,
}

// The generated code:
impl ToJson for S {
    fn to_json(&self) -> serde_json::Value {
        json!({
            "a": ToJson::to_json(&self.a),
            "b": ToJson::to_json(&self.b),
        })
    }
}
```

dtolnay:
Fixes #294.

### DIFF
--- a/json/src/macros.rs
+++ b/json/src/macros.rs
@@ -266,6 +266,6 @@ macro_rules! json_internal {
     // Any Serialize type: numbers, strings, struct literals, variables etc.
     // Must be below every other rule.
     ($other:expr) => {
-        $crate::value::ToJson::to_json(&$other).unwrap()
+        $crate::to_value(&$other).unwrap()
     };
 }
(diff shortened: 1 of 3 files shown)


dtolnay:
Merging because I see a thumbs up in #294 so I assume @oli-obk is on board with this.

Issue #70, opened 2016-05-13
Ability to step past errors in StreamDeserializer if input is valid JSON


dtolnay:
The current behavior of StreamDeserializer is that if it hits an error, that same error is returned in perpetuity which is typically not what you want. (If that *is* what you want, it is easy to implement that behavior atop what I propose below.)

Suppose we have a stream deserializer that expects to read a stream of Strings. If it receives the input `"a" {"b": 0} "c"`, a reasonable thing to expect as output would be `"a" (error) "c"` because the second JSON object failed to deserialize to a String.

On the other hand, if the input is not a valid JSON stream I believe the StreamDeserializer should retain its current behavior because there is no way for it to regain its bearings in an invalid JSON stream. For example if the input is `"a" {"b{{"c" "d"`, I would expect the output to be `"a" (error) (error) (error...)`.

dtolnay:
The desired behavior:

```rust
let data = b"[0] {"k":"v"} [1]";

let de = serde_json::Deserializer::from_slice(data);
let mut stream = de.into_iter::<Vec<i32>>();
assert_eq!(0, stream.byte_offset());

println!("{:?}", stream.next()); // [0]
assert_eq!(3, stream.byte_offset());

println!("{:?}", stream.next()); // type error
assert_eq!(13, stream.byte_offset());

println!("{:?}", stream.next()); // [1]
assert_eq!(17, stream.byte_offset());
```

dbeckwith:
Has there been any progress towards implementing the change mentioned in the `StreamDeserializer` docs?

> _Note_: In the future this method may be changed to return the number of bytes so far deserialized into a successful T _or_ syntactically valid JSON skipped over due to a type error. See serde-rs/json#70 for an example illustrating this.

This is the behavior I was expecting from `StreamDeserializer::byte_offset`; that it would represent the total number of bytes processed so far, regardless of the result those bytes produced. Although it's not clear to me what the behavior should be for syntactically _invalid_ JSON.

dbeckwith:
While this could still be useful to implement, I'll describe my workaround as well which may be a better solution. Rather than using the `StreamDeserializer` to produce values of `T` directly, so it's doing both parsing and deserialization, I'm just getting `Value`s from the `StreamDeserializer` so all it's doing is parsing JSON and will always advance if it finds valid JSON. Then later I'm using `from_value` to convert the JSON into my domain types.

Arnavion:
[...]

PR #303, opened 2017-04-12
Expose number of bytes processed by StreamDeserializer

### LINKED ISSUE #183: Expose number of bytes processed by StreamDeserializer

dtolnay:
Fixes #183.

The byte offset is the number of bytes so far deserialized into a successful `T`. If a stream deserializer returns an EOF error, new data can be joined to `old_data[stream.byte_offset()..]` to try again.

```rust
let data = b"[0] [1] [";

let de = serde_json::Deserializer::from_slice(data);
let mut stream = de.into_iter::<Vec<i32>>();
assert_eq!(0, stream.byte_offset());

println!("{:?}", stream.next()); // [0]
assert_eq!(3, stream.byte_offset());

println!("{:?}", stream.next()); // [1]
assert_eq!(7, stream.byte_offset());

println!("{:?}", stream.next()); // error
assert_eq!(8, stream.byte_offset());

// If err.is_eof(), can join the remaining data to new data and continue.
let remaining = &data[stream.byte_offset()..];
```

cc @vorner @ShepMaster who were interested in #212.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1067,6 +1067,13 @@ impl<'de, R, T> StreamDeserializer<'de, R, T>
     /// // If err.is_eof(), can join the remaining data to new data and continue.
     /// let remaining = &data[stream.byte_offset()..];
     /// ```
+    ///
+    /// *Note:* In the future this method may be changed to return the number of
+    /// bytes so far deserialized into a successful T *or* syntactically valid
+    /// JSON skipped over due to a type error. See [serde-rs/json#70] for an
+    /// example illustrating this.
+    ///
+    /// [serde-rs/json#70]: https://github.com/serde-rs/json/issues/70
     pub fn byte_offset(&self) -> usize {
         self.offset
     }
--- a/src/de.rs
+++ b/src/de.rs
@@ -97,8 +97,10 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     {
         // This cannot be an implementation of std::iter::IntoIterator because
         // we need the caller to choose what T is.
+        let offset = self.read.byte_offset();
         StreamDeserializer {
             de: self,
+            offset: offset,
             output: PhantomData,
             lifetime: PhantomData,
         }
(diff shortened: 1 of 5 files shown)


dtolnay:
I will also add a link to #70 saying that in the future I would like to change this method to return the number of bytes so far deserialized into a successful `T` *or* syntactically valid JSON skipped over due to a type error. I do not consider that a breaking change.

```rust
let data = b"[0] {"k":"v"} [";

let de = serde_json::Deserializer::from_slice(data);
let mut stream = de.into_iter::<Vec<i32>>();
assert_eq!(0, stream.byte_offset());

println!("{:?}", stream.next()); // [0]
assert_eq!(3, stream.byte_offset());

println!("{:?}", stream.next()); // type error
assert_eq!(13, stream.byte_offset());
```

vorner:
As I'm one of the potential users of this, I can probably comment on the interface. I believe this would solve my use case (or make it significantly easier to solve, the `is_eof` was the crucial part and the number of bytes could be extracted by using a custom iterator ‒ possible, but complex and tedious).

One question, though. If I hit the EOF, I can take the remainder of the data, append more to it and create a new deserializer on top of it (obviously). But is there a way to reuse the existing one? Or is the deserializer lightweight enough so the creation of a new one doesn't matter?

dtolnay:
Constructing a deserializer is lightweight enough.

In the future I see this being used in something like [`tokio-serde-json`](https://github.com/carllerche/tokio-serde-json) which would handle the EOF checking and juggling of the deserializer for you. For serde_json I would like to focus on providing the foundational pieces to make something like that possible.

PR #306, opened 2017-04-21
Use the new linked-hash-map entry API

dtolnay:


src/value.rs
```diff
@@ -988,12 +988,7 @@ impl Index for str {
         }
         match *v {
             Value::Object(ref mut map) => {
-                // TODO: use entry() once LinkedHashMap supports entry()
-                // https://github.com/contain-rs/linked-hash-map/issues/5
-                if !map.contains_key(self) {
-                    map.insert(self.to_owned(), Value::Null);
-                }
-                map.get_mut(self).unwrap()
+                map.entry(self.to_owned()).or_insert(Value::Null)
```
oli-obk: Now we unconditionally do a `self._owned()`

dtolnay: Yes and before it unconditionally did two (sometimes three) traversals to find the key. Maybe this should wait until we get https://github.com/rust-lang/rfcs/pull/1769?

oli-obk: Nah, let's merge now. Once entry gets expanded there'll be a clippy lint complaining about it

dtolnay: Ah perfect!

src/value/index.rs
```diff
@@ -80,12 +80,7 @@ impl Index for str {
         }
         match *v {
             Value::Object(ref mut map) => {
-                // TODO: use entry() once LinkedHashMap supports entry()
-                // https://github.com/contain-rs/linked-hash-map/issues/5
-                if !map.contains_key(self) {
-                    map.insert(self.to_owned(), Value::Null);
-                }
-                map.get_mut(self).unwrap()
+                map.entry(self.to_owned()).or_insert(Value::Null)
```
burdges: If I understand, this results in `self.to_owned()` being called when the entry already exists, while previously it was called only when the entry did not exist.

dtolnay: Yes. And previously it did two (sometimes three) traversals to find the key. We are working on a better approach in rust-lang/rfcs#1769 but for now this is the most readable and correct-looking.

PR #307, opened 2017-04-21
Split value into pieces

killercup:
The mother of all diff noise. The only _new_ code is a bunch of `mod` statements.


dtolnay:
Sweet. I filed #308 to look into why the AppVeyor build failed the first time.

PR #311, opened 2017-04-27
Allow integers to be used as map keys again

dtolnay:
Fixes https://github.com/serde-rs/json/pull/177#issuecomment-297780506 by deserializing map keys with a special Deserializer that allows the same types that the map key Serializer allows.

cc @arthurprs

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -988,8 +988,11 @@ macro_rules! deserialize_integer_key {
             self.de.eat_char();
             self.de.str_buf.clear();
             let string = try!(self.de.read.parse_str(&mut self.de.str_buf));
-            let integer = try!(string.parse().map_err(de::Error::custom));
-            visitor.$visit(integer)
+            match (string.parse(), string) {
+                (Ok(integer), _) => visitor.$visit(integer),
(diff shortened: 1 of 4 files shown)


arthurprs:
Working as expected! Thanks for the quick solution and all the work on serde and related libraries.

dtolnay:
@arthurprs I released 1.0.1 with this fix. Sorry about the break! This was originally handled by implicit conversions in Serde, didn't have a unit test in serde_json and I broke it unintentionally in https://github.com/serde-rs/serde/pull/839.

src/de.rs
```diff
@@ -973,6 +973,105 @@ impl<'de, 'a, R: Read<'de> + 'a> de::VariantAccess<'de> for UnitVariantAccess<'a
...
+    de: &'a mut Deserializer<R>,
+}
+
+macro_rules! deserialize_integer_key {
+    ($deserialize:ident => $visit:ident) => {
+        fn $deserialize<V>(self, visitor: V) -> Result<V::Value>
+        where
+            V: de::Visitor<'de>,
+        {
+            self.de.eat_char();
```
bjorn3: Maybe add an assert that this is a `"` char.

Issue #313, opened 2017-05-02
Serialize/Deserialize implementations are very slow to compile


Rufflewind:
I’m currently trying out the [`pandoc_ast`](https://github.com/oli-obk/pandoc-ast) crate, which is contains [a handful of recursive algebraic data types](https://github.com/oli-obk/pandoc-ast/blob/92060b1d3503557376313074cf7bd3033eda8634/src/lib.rs) with the `Serialize` and `Deserialize` traits derived.  When I tried the minimal example below with optimizations on (`--release`), it took a whole minute to compile!

```rust
extern crate pandoc_ast;
fn main() {
    pandoc_ast::filter(String::new(), |p| p);
}
```

I cloned the `pandoc_ast` repo and messed around a bit: it seems that the library itself takes practically no time to compile, but if I add a concrete function that instantiates `serde_json::from_str<Pandoc>`, the compile time spikes to 60 seconds or so.  I'm not sure if this is caused by `serde`, `serde_derive`, or `serde_json`.

For the smaller nonrecursive enums (e.g. `pandoc_ast::MathType`) with only a few variants, it takes about 2-3 seconds to instantiate.  The larger ones (e.g. `pandoc_ast::Block`, which contains many variants as well as `pandoc_ast::Inline`, which itself contains even more variants) take about 50 sec.

Environment:

  - rustc 1.19.0-nightly (777ee2079 2017-05-01)
  - Linux 4.10.13
  - incremental compilation was not used
  - `cargo (build|test) --release`

dtolnay:
Thanks for reporting this! That is definitely substantially longer than expected. Looks like practically all of that time is spent in LLVM:

```rust
$ cargo rustc --release -- -Z time-passes

  time: 1.637; rss: 275MB	translation
  time: 0.000; rss: 275MB	assert dep graph
  time: 0.000; rss: 275MB	serialize dep graph
    time: 1.474; rss: 276MB	llvm function passes [0]
    time: 52.386; rss: 317MB	llvm module passes [0]
    time: 8.498; rss: 318MB	codegen passes [0]
    time: 0.001; rss: 318MB	codegen passes [0]
  time: 62.364; rss: 318MB	LLVM passes
  time: 0.000; rss: 318MB	serialize work products
    time: 0.247; rss: 170MB	running linker
  time: 0.248; rss: 170MB	linking
      Finished release [optimized] target(s) in 64.62 secs
```

I will try to narrow down whether a specific pattern on our end is triggering slow behavior.

[...]

PR #314, opened 2017-05-03
Reduce visitor-dependence in number parsers

Rufflewind:
All these functions end up calling `visit_f64` so rewriting these functions to avoid dependence on the visitor type `V` is a no-brainer:

  - `parse_long_integer`
  - `parse_decimal`
  - `parse_exponent`
  - `parse_exponent_overflow`
  - `visit_f64_from_parts`

Slightly alleviates #314.

---

Benchmarks on the [minimal Pandoc example](https://github.com/serde-rs/json/issues/313#issuecomment-298806623) show a 10% reduction in both build time and memory usage:

- 14.4 s → 13.0 s
- 451 M → 407 M

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -80,6 +80,25 @@ macro_rules! overflow {
     }
 }
 
+enum Number {
+    F64(f64),
+    U64(u64),
+    I64(i64),
+}
+
+impl Number {
+    fn visit<'de, V>(self, visitor: V) -> Result<V::Value>
+    where
+        V: de::Visitor<'de>,
+    {
+        match self {
+            Number::F64(x) => visitor.visit_f64(x),
+            Number::U64(x) => visitor.visit_u64(x),
+            Number::I64(x) => visitor.visit_i64(x),
+        }
+    }
+}
+
 impl<'de, R: Read<'de>> Deserializer<R> {
     /// The `Deserializer::end` method should be called after a value has been fully deserialized.
     /// This allows the `Deserializer` to validate that the input stream is at the end or that it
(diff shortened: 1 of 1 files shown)


Rufflewind:
Some new stats for each commit:

- 8202c68c0a1d6162f3766e072ad154eee78bfde2 12.4 s 448 M
- 476e753db863c7f9059f82c5ae8c6a2220b2fc5c 11.2 s 411 M (Δt = −10%)
- 2b68532e9d3ad68df07c3b4cf221dc9dbaa40f6e 10.4 s 387 M (Δt = −7%)
- 7b9ad82c9562e2805ff40d406b1998ae69200f14 9.6 s 366 M (Δt = −8%)

(I just upgraded Rust nightly, so these stats are a little lower than what I had in the initial post.)

dtolnay:
Thanks! 23% compile time improvement from this is impressive and there is a lot more we can do to improve it further. We definitely need to evaluate the runtime performance impact too though. I don't have time this week to put into benchmarking but I expect to get to it this weekend.

dtolnay:
Benchmark results are in! The first two commits show no performance regression and a consistent 15% improvement in compile time on the [json-benchmark](https://github.com/serde-rs/json-benchmark) repo. The third commit runs about 30% slower across the board and confusingly also compiles 20-30% slower for me than the other commits. So I am merging 476e753db863c7f9059f82c5ae8c6a2220b2fc5c and 2b68532e9d3ad68df07c3b4cf221dc9dbaa40f6e but not 7b9ad82c9562e2805ff40d406b1998ae69200f14. We will need to figure out what went wrong there. Possibly a missing `#[inline]` on one of the functions.

alexbool:
Great! Do we also need to patch other formats, YAML in particular?

dtolnay:
Yes, serde_yaml certainly has spots where non-Visitor-depenent functionality could be factored out.

Issue #319, opened 2017-05-13
Better "invalid type" errors


dtolnay:
```rust
extern crate serde_json;

fn main() {
    println!("{}", serde_json::from_str::<String>("null").unwrap_err());
}
```

> invalid type: unit value, expected a string

The term "unit value" comes from `<serde::de::Unexpected as Display>::fmt`. In JSON this message should refer to "null" instead.

> invalid type: null, expected a string

This will involve overriding `<serde_json::Error as serde::de::Error>::invalid_type`.

dtolnay:
Fixed in #322.

PR #322, opened 2017-05-21
Better invalid type error message

GyrosOfWar:
Addresses #319.  I'm not sure if the change handles all the relevant cases.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -389,6 +389,14 @@ impl de::Error for Error {
             ),
         }
     }
+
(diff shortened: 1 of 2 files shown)


GyrosOfWar:
That's odd.. tests don't fail on my machine. I'll look into it.

dtolnay:
Thanks @GyrosOfWar. Let us know if you get stuck tracking down the problem.

dtolnay:
Nice work.

PR #330, opened 2017-06-01
Implements PartialOrd for Number.

### LINKED ISSUE #328: Impl PartialOrd for Number
It seems possible to implement both of these for Number.

remexre:
Resolves #328.

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -238,29 +238,45 @@ impl Debug for Number {
     }
 }
 
+impl Eq for Number {}
+
+impl Ord for Number {
+    fn cmp(&self, other: &Number) -> Ordering {
+        match (self.n, other.n) {
+            (N::PosInt(a), N::PosInt(b)) => a.cmp(&b),
+            (N::NegInt(a), N::NegInt(b)) => a.cmp(&b),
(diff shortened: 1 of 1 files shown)


src/number.rs
```diff
@@ -237,6 +238,40 @@ impl Debug for Number {
     }
 }
 
+impl PartialOrd for Number {
+    fn partial_cmp(&self, other: &Number) -> Option<Ordering> {
+        fn reverse(o: Ordering) -> Ordering {
```
clarcharr: Why not use `Ordering::reverse` instead?

remexre: Oh, didn't know that was a thing.

remexre:
Okay, fixed the reverse

clarcharr:
You also probably should add tests for this. Also, I have a feeling that we'd still be able to turn this into an `Ord` definition with a bit of work; I'll try and see what I can do.

remexre:
`Ord` required `Eq`, though, and `f64`s are only `PartialEq` (because `NaN`s). Since we check that `NaN` never appears, it is legal to make this type `Eq` and `Ord`. I'll add the tests if this works (unrelated tests seem to be failing locally; it might just be the nightly I'm on, so I'm gonna let Travis check it).

PR #332, opened 2017-06-10
Switch to using char::encode_utf8

### LINKED ISSUE #270: Replace format_escaped_char with a safe implementation
#267 added this implementation:

```rust
let mut buf = [0; 4];
write!(&mut buf[..], "{}", value).unwrap();
let s = unsafe { str::from_utf8_unchecked(&buf[0..value.len_utf8()]) };
format_escaped_str(wr, formatter, s)
```

Once we drop support for <1.15.0, the implementation can be safe:

```rust
format_escaped_str(wr, formatter, value.encode_utf8(&mut [0; 4]))
```

erickt:
Now that Rust 1.18 has come out, and according to our policy of supporting
the past 3 versions of rust, we can finally switch to using
`char::encode_utf8` to encode a character into a string.

Closes #270.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ cache: cargo
 rust:
   - stable
   - nightly
-  - 1.12.0
-  - 1.13.0
+  - 1.16.0
+  - 1.17.0
   - beta
 
 matrix:
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -219,8 +219,9 @@ where
 
     #[inline]
     fn serialize_char(self, value: char) -> Result<()> {
-        try!(format_escaped_char(&mut self.writer, &mut self.formatter, value).map_err(Error::io));
-        Ok(())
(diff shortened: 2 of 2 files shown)


dtolnay:
I added 117cf7ade6591f2620c7b1b61a97e02325cd1b5d to test 1.15.0 as well, just so we notice what eventually breaks it.

apoelstra:
Over a year later this is the still only thing preventing compilation on rustc 1.14, which is the currently shipping rustc on Debian stable (on i386 and arm64 anyway, see https://packages.debian.org/stretch/rustc ).

What do people think about restoring the old behavior behind a cargo feature so that Debian stable users can use the latest serde?

dtolnay:
As idiomatic use of serde_json revolves so heavily around derive, Rust 1.15 is a reasonable minimum version for us to support. Debian stable users are welcome to continue using Serde 0.9 or maintain a fork of Serde 1.0 under a different name with patches to support Rust 1.14.

PR #347, opened 2017-08-22
Add from_async_reader to simplify integration with tokio

mehcode:
I'm working on an asynchronous web framework for Rust, [Salt](https://github.com/mehcode/salt-rs), and one of my current tasks is to make consuming and producing JSON ergonomic.

Currently, I have this:

https://github.com/mehcode/salt-rs/blob/69cd12b27953f8b602023d8884ceede5dc628535/examples/json.rs#L40-L51

```rust
fn index(ctx: Context) -> impl Future<Item = Response, Error = Error> {
    io::read_to_end(ctx, Vec::new()).from_err().and_then(|(_, buffer)| {
        future::done(serde_json::from_slice(&buffer)).from_err()
    }).and_then(|body: RequestBody| {
        future::done(serde_json::to_string(&json!({
            "id": 20u8,
            "name": body.name,
        }))).from_err()
    }).map(|s| {
        Response::with(s).with_header(header::ContentType::json())
    })
}
```

It'd be very convenient if `serde_json` could parse JSON from an `tokio_io::AsyncRead` (in addition to a `Read` as it currently does).

Here's an example of the improvement possible with this PR:

https://github.com/mehcode/salt-rs/blob/ad4eaada4107f6f496f722a906d670a4afa3a97c/examples/json.rs#L22-L31

```rust
fn index<'a>(ctx: Context) -> impl Future<Item = Response> + 'a {
    serde_json::from_async_reader(ctx).and_then(|(_, body): (_, RequestBody)| {
        future::done(serde_json::to_string(&json!({
            "id": 20u8,
            "name": body.name,
        })))
    }).map(|s| {
        Response::with(s).with_header(header::ContentType::json())
    })
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ num-traits = "0.1.32"
 linked-hash-map = { version = "0.4.1", optional = true }
 itoa = "0.3"
 dtoa = "0.4"
+futures = { version = "0.1.14", optional = true }
+tokio-io = { version = "0.1.2", optional = true }
 
 [dev-dependencies]
 compiletest_rs = "0.2"
@@ -37,3 +39,6 @@ default = []
 # This allows data to be read into a Value and written back to a JSON string
 # while preserving the order of map keys in the input.
 preserve_order = ["linked-hash-map"]
+
+# Integrate with tokio-io to add serde_json::from_async_reader
+tokio = ["futures", "tokio-io"]
--- a/src/de.rs
+++ b/src/de.rs
@@ -18,6 +18,9 @@ use super::error::{Error, ErrorCode, Result};
 
 use read::{self, Reference};
 
+#[cfg(feature = "tokio")]
+use futures::{Future, future};
+
 pub use read::{Read, IoRead, SliceRead, StrRead};
 
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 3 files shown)


mehcode:
@dtolnay Sorry it took so long to respond. I did see your messages on IRC.

---

If we consider an ideal future for tokio ..

```rust
#[async]
fn index(ctx: Context) -> Result<Response> {
    let (_, buffer) = await!(io::read_to_end(ctx, Vec::new()))?;
    let body: RequestBody = serde_json::from_slice(&buffer)?;

    // [...]
}
```

That's still not good enough. We have `serde_json::from_reader` for dealing directly with `io::Read`.

```rust
fn index(ctx: Context) -> Result<Response> {
    // If we pretend that `Context` is not `AsyncRead`
    let body: RequestBody = serde_json::from_reader(ctx)?;

    // [...]
}
```

As `tokio::AsyncRead` is `std::io::Read` I'm actually hoping specialization can allow `serde_json::from_reader` to transparently work on `AsyncRead`.


```rust
#[async]
fn index(ctx: Context) -> Result<Response> {
    let body: RequestBody = await!(serde_json::from_reader(ctx))?;

    // [...]
}
```

That is my ideal here. I understand that as serde-json is `1.x` it can't easily depend on `0.x` dependencies.

Perhaps when tokio is at the point of enabling the ideal we could add a `unstable` or `nightly` feature flag here that would allow it to be added here without compromising the `1.x`?

dtolnay:
In line with those ideas, a blanket impl in an external library could look like this. What do you think?

```rust
#[async]
fn index(ctx: Context) -> Result<Response> {
    let body: RequestBody = await!(ctx.map_bytes(serde_json::from_slice))?;

    // [...]
}
```

I think the tokio ecosystem has room for a high-quality library for composing async read and write with other operations, sort of like what byteorder's extension traits do for ordinary read and write.

PR #353, opened 2017-09-01
change error type for trailing commas, fixes #352

### LINKED ISSUE #352: misleading error message
I was going through the [serde_json :: de :: from_reader](https://docs.serde.rs/serde_json/de/fn.from_reader.html) example and encountered an error at run time "KeyMustBeAString". I went back to my json file and looked. All of the keys were strings... I scratched my head and stared. Then I noticed that I had a trailing comma after the last value.

It seems like adding another error code that has something to do with trailing commas would make the error messages more human friendly.

test.json
```json
{
  "fingerprint": "gray",
  "location": "1",
}
```

error message:
```rust
ErrorImpl { 
    code: KeyMustBeAString,
    line: 4,
    column: 1
 }
```

fmoor:
**NOTE:** There is one small defect with this patch. The line and column numbers that the error message reports match the position of the closing brace after the trailing comma and not the position of the trailing comma.

I plant to continue searching for the logic that determines the line and column numbers. If someone wants to speed things up and point me towards the relevant code it would be much appreciated.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -105,6 +105,7 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     /// only has trailing whitespace.
     pub fn end(&mut self) -> Result<()> {
         match try!(self.parse_whitespace()) {
+            Some(b',') => Err(self.peek_error(ErrorCode::TrailingComma)),
             Some(_) => Err(self.peek_error(ErrorCode::TrailingCharacters)),
             None => Ok(()),
         }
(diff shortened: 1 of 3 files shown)


fmoor:
There is still the problem of reporting misleading line and column numbers.

fmoor:
@dtolnay why are there suddenly a bunch of clippy errors in code I haven't touched? Should I make all of the suggested changes?

dtolnay:
Clippy adds new lints occasionally. I filed https://github.com/serde-rs/json/issues/354 to follow up.

PR #358, opened 2017-09-07
Resolves #334 - Extend Serializer's default recursion limit to 255

### LINKED ISSUE #334: Allow increasing recursion limit

Yoric:
By opposition to https://github.com/serde-rs/json/pull/346, this doesn't break the stack.

Also, as discussed in https://github.com/serde-rs/json/pull/346, `remaining_depth` becomes a `usize`.


Yoric:
> Would it be possible for you to use a fork of serde_json until we come up with the right API for controlling recursion limit?

I guess.

PR #369, opened 2017-10-12
Added NumberKind

### LINKED ISSUE #366: Exposing Number Kind
I have been working on a json-to-struct generator for a project, and I ran into an issue that I did not actually know the type of number. It would be useful to expose the type of the Number.

For reference: https://github.com/serde-rs/json/blob/master/src/number.rs#L17-L32

After discussion with @oli-obk we came to the following possibilities:

1. Making `Number::n` and `N` pub
https://play.rust-lang.org/?gist=76af67ced7d5a9ff3aad1d0a10878b46&version=stable
2. Introducing a `NumberKind` type and implementing a `kind()` function on `Number`
https://play.rust-lang.org/?gist=017cb7efe26cd6bf9a2b6b35f0f76dce&version=stable
3. Merging `N` into `Number`, making `Number` the enum itself
https://play.rust-lang.org/?gist=ee1d87a959c1e3d8da982418222b65ac&version=stable

VictorKoenders:
Closes #366

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -31,15 +31,14 @@ enum N {
     Float(f64),
 }
 
-/// Describes the kind of number we're dealing with
-/// This is a hint about how the number is stored internally
+/// Gives a hint to the value of the Number.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum NumberKind {
-    /// The number is a positive integer, and can be stored in an u64
+    /// The number is a positive integer
     PosInt,
-    /// The number is a negative integer, and can be stored in an i64
+    /// The number is a negative integer
     NegInt,
-    /// The number is a float, and can be stored in an f64
+    /// The number is a float
     Float,
 }
 
@@ -232,7 +231,7 @@ impl Number {
         }
     }
     
-    /// Gets a hint on how the number is stored internally.
+    /// Gives a hint to the value of the Number.
     ///
     /// ```rust
     /// # use serde_json::{Number, NumberKind};
--- a/src/number.rs
+++ b/src/number.rs
@@ -235,9 +235,12 @@ impl Number {
     /// Gets a hint on how the number is stored internally.
     ///
     /// ```rust
-    /// assert_eq!(Number::from_f64(256.0).kind(), NumberKind::Float)
-    /// assert_eq!(Number::from(256).kind(), NumberKind::PosInt)
-    /// assert_eq!(Number::from(-256).kind(), NumberKind::NegInt)
+    /// # use serde_json::{Number, NumberKind};
+    /// #
+    /// assert_eq!(Number::from_f64(256.0).unwrap().kind(), NumberKind::Float);
+    /// assert_eq!(Number::from(256).kind(), NumberKind::PosInt);
+    /// assert_eq!(Number::from(-256).kind(), NumberKind::NegInt);
+    /// ```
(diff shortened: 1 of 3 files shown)


src/number.rs
```diff
@@ -31,6 +31,18 @@ enum N {
     Float(f64),
 }
 
+/// Describes the kind of number we're dealing with
+/// This is a hint about how the number is stored internally
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum NumberKind {
+    /// The number is a positive integer, and can be stored in an u64
```
dtolnay: I would prefer not to make the guarantee that all positive integers can be stored in u64. If we move to an arbitrary precision representation, something can be a positive integer without fitting in u64. Same for the negative integer and float cases.

src/number.rs
```diff
@@ -31,6 +31,18 @@ enum N {
     Float(f64),
 }
 
+/// Describes the kind of number we're dealing with
+/// This is a hint about how the number is stored internally
```
dtolnay: This needs punctuation and a newline in between, or else rustdoc sees it as one sentence.

---

selection_071

src/number.rs
```diff
@@ -219,6 +231,23 @@ impl Number {
             None
         }
     }
+    
+    /// Gets a hint on how the number is stored internally.
```
dtolnay: I don't think we should conceptualize this as a hint on how the number is stored internally. The internal storage is not the user's business. Could you frame this in a way that does not involve internal storage?

VictorKoenders:
For my specific use case I want to know the minimal data type that's required to store the number that's stored, so that would work.

I would still like to return a (bitflagged?) enum. That would leave the functionality together and you don't end up having a bunch of random functions on the Number struct that are not relevant to most people.

What do you think of this?

```Rust
bitflags! {
    struct NumberProperties: u8 {
        const PROP_NEGATIVE = 0b0000_0001;
        const PROP_DECIMAL  = 0b0000_0010;
    }
}

impl NumberProperties {
    pub fn is_positive(&self) -> bool { !self.is_negative() }
    pub fn is_negative(&self) -> bool { self.contains(NumberProperties::PROP_NEGATIVE) }
    
    pub fn is_integer(&self) -> bool { !self.is_integer() }
    pub fn is_decimal(&self) -> bool { self.contains(NumberProperties::PROP_DECIMAL) }
}
```

dtolnay:
I would prefer to write `n.is_integer()` over `n.properties().is_integer()`. I think given that f64 and friends don't see a need for properties(), we should not need it either. These are as relevant to anyone using Number as any of the other methods on Number.

oli-obk:
How about we expose a `try_to_primitive` method that produces the enum we currently use as a representation? (Well an `Option` to it). This will keep working in the future, unrepresentable stuff simply gets turned into `None`

oli-obk:
@dtolnay what do you think about the single method for converting to the enum?

dtolnay:
I would be on board with a single method to try representing a Number as a u64/i64/f64 primitive. Good idea!

You should feel free to design the returned enum from scratch without trying to mirror the current `N` internal representation. In particular, PosInt is a confusing name because the integer 0 is not positive.

dtolnay:
I am closing this PR because it has gone inactive, but I would be happy to consider https://github.com/serde-rs/json/pull/369#issuecomment-338911168 in a new PR.

PR #384, opened 2017-11-24
Support top-level primitives in StreamDeserializer

bouk:
Currently StreamDeserializer only supports objects and arrays, because it couldn't handle things like `truefalse` properly before. This PR changes the behavior of the parser so that it checks after parsing an identifier whether there's any trailing characters that aren't one of ` \n\t\r"[]{},:` or EOF, and if there is it will give an error. This means it will error immediately when parsing `truefalse` instead of first emitting `true` followed by an error.

Related: https://github.com/serde-rs/json/pull/301

cc @dtolnay

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1302,9 +1302,8 @@ where
 /// A stream deserializer can be created from any JSON deserializer using the
 /// `Deserializer::into_iter` method.
 ///
-/// The data must consist of JSON arrays and JSON objects optionally separated
-/// by whitespace. A null, boolean, number, or string at the top level are all
-/// errors.
+/// The data can consist of any JSON value. Values need to be a self-delineating value e.g.
+/// arrays, objects, or strings, or be followed by whitespace or a self-delineating value.
 ///
 /// ```rust
 /// extern crate serde_json;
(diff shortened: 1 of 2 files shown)


tests/test.rs
```diff
@@ -669,8 +669,8 @@ fn test_parse_bool() {
             ("truz", "expected ident at line 1 column 4"),
             ("f", "expected ident at line 1 column 1"),
             ("faz", "expected ident at line 1 column 3"),
-            ("truea", "trailing characters at line 1 column 5"),
-            ("falsea", "trailing characters at line 1 column 6"),
```
bouk: This PR changes these error messages. Let me know if that's a problem, although in my opinion the new error is more accurate

bouk:
Thanks for the reply!

* I am creating a tool that ingests a stream of JSON values, transforms them, and outputs them again. I am just using this crate as a JSON parser for this purpose, not using any of Serde's features. I foresee some use cases that would involve any kind of JSON value, so I figured I would take a stab at implementing this feature
* will do
* Yes, I just realized this is totally possible by moving the end of value check into the next() function of StreamDeserializer. Will make that change

I will make the required changes later this week

dtolnay:
Sounds good. I am on board with this change once the documentation is updated and the logic moves into StreamDeserializer. Thanks!

bouk:
@dtolnay I've updated the documentation and behavior, let me know what you think. Open to suggestions for the docs, it came out a bit awkward (but I'm not sure how else to describe it)

dtolnay:
I published serde_json 1.0.7 with this change.

Issue #405, opened 2018-01-20
Avoid unnecessary allocation in `<Value as From<Cow<'a, str>>>::from()`


H2CO3:
The current implementation of `From<Cow<'a, str>>::from()` is:

```rust
fn from(f: Cow<'a, str>) -> Self {
    Value::String(f.to_string())
}
```
Now, `Cow<'a, str>::to_string()` unconditionally allocates and copies (and it has to, because it takes `&self` by pointer, as dictated by the `ToString` trait). But since `From::from` takes its argument by value, I believe it would be possible to implement it like this instead:

```rust
fn from(f: Cow<'a, str>) -> Self {
    match f {
        Cow::Borrowed(slice) => Value::String(slice.to_owned()),
        Cow::Owned(buf) => Value::String(buf),
    }
}
```

What do you think? If it's possible and worth implementing, I'd be happy to open a PR.

dtolnay:
Seems reasonable! Please send a PR.

dtolnay:
`Value::String(f.into_owned())` should work.

H2CO3:
Oh, indeed, good catch!

PR #406, opened 2018-01-21
Remove copy and cleanup

H2CO3:
Fixes https://github.com/serde-rs/json/issues/405. Bonus: a little cleanup/code simplification in the same file.


H2CO3:
You're welcome! 👍

Issue #364, opened 2017-10-05
Inappropriate "invalid unicode code point" message


saarw:
In the following test I get a InvalidUnicodeCodePoint error, but aren't Rust string literals guaranteed to be valid unicode sequences? (says so here https://doc.rust-lang.org/1.7.0/book/strings.html )
```
extern crate serde;
extern crate serde_json;

#[macro_use]
extern crate serde_derive;

#[test]
fn test_json() {
    #[derive(Debug, Deserialize)]
    struct A {
        name: String,
        value: String
    };
    let val = r#"{
        "name": "a",
        "value": "<img src=javascript:
        sneaky>"
    }"#;
    let result = ::serde_json::from_slice::<A>(val.as_bytes());
    println!("{:?}", result);
    assert!(result.is_ok());
}
```

dtolnay:
JSON does not allow newlines in a string, so this is an error. But you are right that the error message you get is not the right one. Would you be interested in submitting a PR to fix the message?

dtolnay:
Fixed in #414.

PR #414, opened 2018-02-26
Improve error when json strings contain control characters

boxofrox:
A first pass at resolving #364.  Comments are welcome.

This patch checks specifically for control characters (ASCII 0x00 - 0x1F) in a JSON string and reports a more relevant error message:

```
control character (\u0000-\u001F) found while parsing a string at line 1 column 1"
```

Points of concern:

- There may be a better (i.e. shorter) error name than `ErrorCode::ControlCharacterWhileParsingString`.
- User confusion with error message.  Will a user suspect that escaped control characters caused the error because I mention the ASCII range formatted as escaped control characters?  Perhaps instead, I should say:

  - `control character (ASCII 0x00 - 0x1F) found while parsing a string at line 1 column 1`, or
  - `unescaped control character (\u0000-\u001F) found while parsing a string at line 1 column 1`, or
  - `raw control character (\u0000-\u001F) found while parsing a string at line 1 column 1`, or
  - some combination of the above.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -73,6 +73,7 @@ impl Error {
             ErrorCode::InvalidNumber |
             ErrorCode::NumberOutOfRange |
             ErrorCode::InvalidUnicodeCodePoint |
+            ErrorCode::ControlCharacterWhileParsingString |
             ErrorCode::KeyMustBeAString |
             ErrorCode::LoneLeadingSurrogateInHexEscape |
             ErrorCode::TrailingComma |
@@ -240,6 +241,9 @@ pub enum ErrorCode {
     /// Invalid unicode code point.
     InvalidUnicodeCodePoint,
 
+    /// Control character found while parsing a string.
+    ControlCharacterWhileParsingString,
+
     /// Object key is not a string.
     KeyMustBeAString,
 
@@ -324,6 +328,9 @@ impl Display for ErrorCode {
             ErrorCode::InvalidNumber => f.write_str("invalid number"),
             ErrorCode::NumberOutOfRange => f.write_str("number out of range"),
             ErrorCode::InvalidUnicodeCodePoint => f.write_str("invalid unicode code point"),
+            ErrorCode::ControlCharacterWhileParsingString => {
+                f.write_str("control character (\\u0000-\\u001F) found while parsing a string")
+            }
             ErrorCode::KeyMustBeAString => f.write_str("key must be a string"),
             ErrorCode::LoneLeadingSurrogateInHexEscape => {
                 f.write_str("lone leading surrogate in hex escape")
--- a/src/read.rs
+++ b/src/read.rs
@@ -181,7 +181,7 @@ where
                 }
                 _ => {
                     if validate {
-                        return error(self, ErrorCode::InvalidUnicodeCodePoint);
+                        return error(self, ErrorCode::ControlCharacterWhileParsingString);
                     }
                     scratch.push(ch);
                 }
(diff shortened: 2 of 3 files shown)


src/error.rs
```diff
@@ -189,6 +190,9 @@ struct ErrorImpl {
 #[doc(hidden)]
 #[derive(Debug)]
 pub enum ErrorCode {
+    /// Control character found while parsing a string.
+    ControlCharacterWhileParsingString,
```
dtolnay: Please move this next to the variant you think is most similar to it -- maybe the unicode one.

src/error.rs
```diff
@@ -58,6 +58,7 @@ impl Error {
         match self.err.code {
             ErrorCode::Message(_) => Category::Data,
             ErrorCode::Io(_) => Category::Io,
+            ErrorCode::ControlCharacterWhileParsingString |
```
dtolnay: This error is a syntax error so it should be categorized as `Category::Syntax` not `Category::Eof`.

src/error.rs
```diff
@@ -307,6 +311,9 @@ impl Error {
 impl Display for ErrorCode {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
+            ErrorCode::ControlCharacterWhileParsingString => {
```
dtolnay: Keep this in the same order as in the `ErrorCode` enum.

src/read.rs
```diff
@@ -358,7 +366,11 @@ impl<'a> SliceRead<'a> {
                 }
                 _ => {
                     if validate {
-                        return error(self, ErrorCode::InvalidUnicodeCodePoint);
+                        if self.slice[self.index] <= b'\x1F' {
+                            return error(self, ErrorCode::ControlCharacterWhileParsingString);
+                        } else {
+                            return error(self, ErrorCode::InvalidUnicodeCodePoint);
```
dtolnay: What input would hit this line?

src/error.rs
boxofrox: Done.  Moved just after `InvalidUnicodeCodePoint`.

src/error.rs
boxofrox: Oops, thought I had that in the syntax category.  Done.

src/error.rs
boxofrox: Done.

src/read.rs
boxofrox: Hmm... I believe `&[b'"', b'\n', b'"']` will hit that line.  I added two entries in `test_parse_slice_err` for this.

Beyond that, I don't really understand when `SliceRead` would be used instead of `IoRead`.  I updated all sites with `InvalidUnicodeCodePoint` except those sites dealing with decoding escaped unicode code points.  I may have gone overboard, so if this change isn't appropriate, I'll take it out.

src/error.rs
dtolnay: Doesn't look like this has been moved yet.

src/read.rs
dtolnay: I mean specifically this line. The new test cases are hitting the return two lines up from this one.

boxofrox: Oh, good catch. I didn't realize the `_` arm of the `match` was only handling the `CT` escapes, since backslash and quote were handled by the other arms.  No input hits that line.  Not there, and not for the other 3 sites I changed.

I originally assumed there were other unicode-related cases that would trigger the `_` arm and threw in the `if-else` to catch those.

Patch simplified to remove the extraneous `if-else`.

dtolnay:
I published serde_json 1.0.10 with this fix.

boxofrox:
Thanks for reviewing!

PR #416, opened 2018-03-02
Arbitrary-precision numerics support

alexreg:
Added support for arbitrary-precision numerics, in a similar way that the *toml* crate does for date-times (using an internal special struct).


alexreg:
@dtolnay The main outstanding issue is the following test failure when the feature is on. (All tests pass when the feature is off.)

```rust
---- src/value/mod.rs - value::Value::is_f64 (line 598) stdout ----
	thread 'rustc' panicked at 'test executable failed:

thread 'main' panicked at 'assertion failed: !v["b"].is_f64()', src/value/mod.rs:11:1
note: Run with `RUST_BACKTRACE=1` for a backtrace.

', librustdoc/test.rs:330:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

This is of course due to the fact that the old logic for `as_f64` does return `Some` for a `PosInt` or `NegInt` value, but the `parse` logic for arbitrary precision numerics does. (I presume you remember discussing this briefly on IRC.) So, we can either add some hacky logic to check to check if a string is "float-like" (this would be more than just checking for a decimal point, because of exponential notation), or break backwards compatibility. What do you think?

Once this is done, and I can do a quick rebase on master, I think we're ready for a proper full review.

alexreg:
Okay, I decided to resolve that in a backwards-compatible manner, in the simplest way I thought possible.

Ready for review now, @dtolnay!

dtolnay:
I agree with the decision to conservatively support `is_f64` as close to backward compatibly as possible.

alexreg:
> Thanks! I began looking through this and everything looks great so far. I continue to be on board with this approach. Will try to get through the rest by ~midweek.
> 
> Could you check on why the Rust 1.15.0 build is failing in Travis?

Okay, glad to hear!

dtolnay:
This one?

```
error: redundant closure found
    --> src/ser.rs:1554:52
     |
1554 |         writer.write_all(value.as_bytes()).map_err(|e| Error::io(e))
     |                                                    ^^^^^^^^^^^^^^^^ help: remove closure as shown: `Error::io`
     |
     = note: #[deny(redundant_closure)] implied by #[deny(clippy)]
     = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.187/index.html#redundant_closure
```

We should follow what it says and use `map_err(Error::io)`.

The 1.15.0 build is going to need another look.

src/number.rs
```diff
@@ -292,90 +390,412 @@ impl<'de> Deserialize<'de> for Number {
...
+    {
+        self.deserialize_any(visitor)
+    }
+
+    #[cfg(feature = "arbitrary_precision")]
+    fn deserialize_str<V>(self, visitor: V) -> Result<V::Value, Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        visitor.visit_str(&self.n)
```
dtolnay: Is this an intentional behavior change or required somewhere else in the PR? The following fails without `arbitrary_precision` but gives you a string with `arbitrary_precision`.

```rust
println!("{:?}", String::deserialize(Number::from(0)).unwrap());
```

We have tended to stay away from such implicit type conversions because they make other functionality fail in surprising ways, for example deserializing to the following untagged enum would produce the string variant even though the input is a number.

```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum E {
    S(String),
    N(i64),
}
```

src/number.rs
```diff
@@ -292,90 +390,412 @@ impl<'de> Deserialize<'de> for Number {
...
-        bool i8 i16 i32 i64 u8 u16 u32 u64 f32 f64 char str string bytes
-        byte_buf option unit unit_struct newtype_struct seq tuple
-        tuple_struct map struct enum identifier ignored_any
+        bool char bytes byte_buf option unit unit_struct newtype_struct seq
+        tuple tuple_struct map struct enum identifier ignored_any
     }
 }
 
-impl<'de, 'a> Deserializer<'de> for &'a Number {
+impl<'de, 'a: 'de> Deserializer<'de> for &'a Number {
```
dtolnay: Is this required? Looks like a breaking change.

alexreg:
> We should follow what it says and use map_err(Error::io).
> 
> The 1.15.0 build is going to need another look.

Oh, I misunderstood that hint. It shouldn't say "remove closure", but rather "change closure to"... anything, that's an issue for the clippy people.

src/number.rs
```diff
@@ -116,10 +151,20 @@ impl Number {
...
+        #[cfg(not(feature = "arbitrary_precision"))]
         match self.n {
             N::Float(_) => true,
             N::PosInt(_) | N::NegInt(_) => false,
         }
+        #[cfg(feature = "arbitrary_precision")]
+        {
+            for c in self.n.chars() {
+                if c == '.' || c == 'e' || c == 'E' {
+                    return true;
```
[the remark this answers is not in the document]
alexreg: Done.

src/number.rs
```diff
@@ -211,24 +265,49 @@ impl Number {
...
     pub fn from_f64(f: f64) -> Option<Number> {
         if f.is_finite() {
-            Some(Number { n: N::Float(f) })
+            let n = {
+                #[cfg(not(feature = "arbitrary_precision"))]
+                { N::Float(f) }
+                #[cfg(feature = "arbitrary_precision")]
+                {
+                    let mut buf = Vec::new();
+                    dtoa::write(&mut buf, f).ok()?;
```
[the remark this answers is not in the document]
alexreg: Fair point.

src/number.rs
alexreg: Sadly yes. This is the most liberal bound that allows for it to compile. We could make the bound dependent on the feature, I suppose, but that might cause confusion for users.

alexreg:
Okay, I've pushed a new revision, which should resolve your above concerns. I suspect everything is in order now, but I'm getting some [strange errors on testing](https://gist.github.com/167798b1e9fb78acf32a6f1258064c09) now, even with the old version (previously confirmed working).

src/number.rs
alexreg: Ah. It was so long ago I originally wrote this, I don't fully remember my motivations, but I think it was mainly for convenience. Your point is well taken though.

src/number.rs
dtolnay: Supposing that we cannot make this breaking change (whether or not it depends on arbitrary_precision), is there any other way to write this impl?

dtolnay:
The "unnecessary parentheses" warnings should go away with a `cargo update`. The "expected expression" failures are from a change in the nightly compiler's diagnostic output and are fixed if you rebase.

alexreg:
Okay, that worked. (Had to clobber the `target/` directories too, which was an issue.)

Thanks for the review. LTGM? 😀

src/number.rs
```diff
@@ -292,90 +391,381 @@ impl<'de> Deserialize<'de> for Number {
...
             N::Float(f) => visitor.visit_f64(f),
         }
     }
 
+    #[cfg(feature = "arbitrary_precision")]
+    #[inline]
+    fn deserialize_any<V>(self, visitor: V) -> Result<V::Value, Error>
+        where V: Visitor<'de>
+    {
+        visitor.visit_map(NumberDeserializer {
```
dtolnay: I think we need to be somewhat more cautious here to avoid breaking existing code. For example in the untagged enum case which relies on deserialize_any:

```rust
#[derive(Deserialize, Debug)]
#[serde(untagged)]
enum E {
    N(i64),
}

fn main() {
    println!("{:?}", E::deserialize(Number::from(0)).unwrap());
}
```

One approach would be to cascade a sequence of checks: if `self.n` can parse as u64, use visit_u64; if `self.n` can parse as i64, use visit_i64; if `self.n` parsed to f64 and then to_string'd gives you the same string as `self.n`, use visit_f64; only otherwise use visit_map.

That way for the most part only code that relies on large numbers (which previously failed to deserialize into serde_json::Number) would be affected.

src/number.rs
dtolnay: This is going to need to be fixed. We cannot release with this breaking change.

src/number.rs
```diff
@@ -292,90 +391,381 @@ impl<'de> Deserialize<'de> for Number {
...
```
dtolnay: Same comment as `<Number as Deserializer<'de>>::deserialize_any`.

src/number.rs
dtolnay: Please remove the methods that are resulting in a behavior change.

src/number.rs
```diff
@@ -292,90 +391,381 @@ impl<'de> Deserialize<'de> for Number {
...
-        $(
-            impl From<$signed_ty> for Number {
-                #[inline]
-                fn from(i: $signed_ty) -> Self {
-                    if i < 0 {
-                        Number { n: N::NegInt(i as i64) }
-                    } else {
-                        Number { n: N::PosInt(i as u64) }
-                    }
+impl FromPrimitive for Number {
```
dtolnay: The num-traits crate is not stable yet so we cannot have these trait impls in the serde_json public API. Please find a different way to implement this.

src/number.rs
```diff
@@ -292,90 +391,381 @@ impl<'de> Deserialize<'de> for Number {
...
+        Some(Number { n: n })
+    }
+
+    #[inline]
+    fn from_f64(n: f64) -> Option<Self> {
+        Self::from_f64(n)
+    }
+}
+
+impl ToPrimitive for Number {
```
dtolnay: Ditto num-traits is not stable yet.

src/number.rs
alexreg: Could we just hide the impl's for them?

alexreg:
I don't see any way to get rid of the `'a: 'de` bound above... do you?

src/number.rs
dtolnay: I would prefer to not have these impls in the public API even if we hide them. Is it possible to implement a different way?

dtolnay:
Yes, the NumberDeserializer struct should not need a lifetime in the first place because the implementation eventually calls to_owned() to make an owned string anyway. If you implement `struct NumberDeserializer` without a lifetime and have it store `Option<String>`, you no longer need to make that breaking change.

src/number.rs
alexreg: Yes, although `FromPrimitive`/`ToPrimitive` does the heavy lifting for us, which is quite nice. I also don't see what's "unstable" about them... is the fact the version is 0.2 really that bad? Serde was 0.x not so long ago, after all.

alexreg:
Ah, fair point that.

src/number.rs
```diff
@@ -292,90 +391,381 @@ impl<'de> Deserialize<'de> for Number {
...
+
+            #[cfg(feature = "arbitrary_precision")]
+            #[inline]
+            fn visit_map<V>(self, mut visitor: V) -> Result<Number, V::Error>
+            where
+                V: de::MapAccess<'de>
+            {
+                let value = visitor.next_key::<NumberKey>()?;
+                if value.is_none() {
+                    return Err(de::Error::custom("number key was not found"))
```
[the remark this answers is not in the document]
alexreg: Yep, sounds like a good idea.

src/number.rs
alexreg: Hmm, is there something special about the way untagged enums work that causes this behaviour? It doesn't seem like the most elegant solution, but I'll have a go at this fix.

alexreg:
@dtolnay All should be in order now, except for the issue with `num-traits`. I await your response re. my above comment. :-)

src/number.rs
dtolnay: I guess I don't understand what heavy lifting you are referring to. Just looking at the [current implementation](https://github.com/serde-rs/json/blob/v1.0.10/src/number.rs#L350) of the `From` impls it looks like making them work in the arbitrary_precision case should be not much more than 1 line of code: `Number { n: i.to_string() }`.

src/number.rs
```diff
@@ -292,90 +389,352 @@ impl<'de> Deserialize<'de> for Number {
...
+
+    #[inline]
+    fn from_f64(n: f64) -> Option<Self> {
+        Self::from_f64(n)
+    }
 }
 
-macro_rules! from_unsigned {
-    ($($unsigned_ty:ident)*) => {
+impl ToPrimitive for Number {
```
dtolnay: Is the ToPrimitive impl needed by any other code in the PR or is it here for symmetry with FromPrimitive?

src/value/de.rs
```diff
@@ -304,10 +411,34 @@ impl<'de> serde::Deserializer<'de> for Value {
...
+    {
+        self.deserialize_any(visitor)
+    }
+
+    #[cfg(feature = "arbitrary_precision")]
+    fn deserialize_struct<V>(self, name: &'static str, fields: &'static [&'static str],
+                             visitor: V) -> Result<V::Value, Error>
+        where V: Visitor<'de>,
+    {
+        if name == SERDE_STRUCT_NAME && fields == &[SERDE_STRUCT_FIELD_NAME] {
```
dtolnay: I don't think we need this case here because the Deserialize impls for Number and Value both call deserialize_any, not deserialize_struct (which is fine). AIUI nothing would ever hit this case.

src/de.rs
```diff
@@ -240,7 +254,7 @@ impl<'de, R: Read<'de>> Deserializer<R> {
         self.fix_position(err)
     }
 
-    fn deserialize_number<V>(&mut self, visitor: V) -> Result<V::Value>
+    fn deserialize_number<V>(&mut self, visitor: V, prim: bool) -> Result<V::Value>
```
dtolnay: I think this method is only called when deserializing a primitive so it shouldn't need the new prim parameter.

Cargo.toml
```diff
@@ -37,3 +37,5 @@ default = []
 # This allows data to be read into a Value and written back to a JSON string
 # while preserving the order of map keys in the input.
 preserve_order = ["linked-hash-map"]
+
+arbitrary_precision = []
```
dtolnay: Please add a comment explaining what is (and maybe what is not) guaranteed when you enable this feature.

src/de.rs
alexreg: I'll double-check, but I don't think that's true. (Well, it wasn't when I originally wrote the code; that may have changed.)

Cargo.toml
alexreg: Something like the following?

src/de.rs
```diff
@@ -1310,13 +1421,23 @@ impl<'de, 'a, R: Read<'de>> de::Deserializer<'de> for &'a mut Deserializer<R> {
...
+        name: &'static str,
+        fields: &'static [&'static str],
         visitor: V
     ) -> Result<V::Value>
     where
         V: de::Visitor<'de>,
     {
+        #[cfg(feature = "arbitrary_precision")]
+        {
+            if name == SERDE_STRUCT_NAME && fields == &[SERDE_STRUCT_FIELD_NAME] {
```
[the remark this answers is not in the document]
alexreg: Not 100% sure, but I think you're right. Removed now.

src/de.rs
alexreg: That wasn't true before, but now that I've removed the block you suggested above in `deserialize_struct`, it should be fine to get rid of.

src/de.rs
```diff
@@ -184,7 +198,7 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     }
 
     #[cold]
-    fn peek_invalid_type(&mut self, exp: &Expected) -> Error {
+    fn peek_invalid_type(&mut self, exp: &Expected, prim: bool) -> Error {
```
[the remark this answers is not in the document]
alexreg: Yep, an oversight of mine. Gone now.

src/number.rs
alexreg: Mainly for symmetry, and because I think it would be nice for consumers. I really don't see the problem with implementing `FromPrimitive` and `ToPrimitive`... people who use `num-traits` will appreciate it, in fact.

src/number.rs
```diff
@@ -292,90 +389,352 @@ impl<'de> Deserialize<'de> for Number {
...
         }
+        self.visited = true;
+        seed.deserialize(NumberFieldDeserializer).map(Some)
+    }
+
+    fn next_value_seed<V>(&mut self, seed: V) -> Result<V::Value, Error>
+    where
+        V: de::DeserializeSeed<'de>,
+    {
+        seed.deserialize(self.number.to_owned().take().unwrap().into_deserializer())
```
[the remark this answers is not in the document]
alexreg: Yep, the `to_owned()` call was just a relic from the old code using a `Cow` value.

alexreg:
@dtolnay The above review should be resolved now, modulo one comment.

dtolnay:
You can't fool me. :stuck_out_tongue_winking_eye: There are 4 comments still visible in https://github.com/serde-rs/json/pull/416/files that all need to be addressed. I will respond to them individually. You need to expand all the files and search for dtolnay.

dtolnay:
I still need to review the serialization side -- src/ser.rs and src/value/ser.rs -- as well as the `fn scan_...` part of src/de.rs.

tests/test.rs
```diff
@@ -795,59 +812,64 @@ fn test_parse_negative_zero() {
...
-           000000000000000000000000000000000000000000000000000000000000\
-           000000000000000000000000000000000000000000000000000000000000\
-           000000000000000000000000000000000000000000000000000000000000\
-           000000000000000000000000000000000000000000000000000000000000\
-           000000000000000000e-10", 1e308),
-    ],
-    );
+    ]);
+
+    if cfg!(not(feature = "arbitrary_precision")) {
```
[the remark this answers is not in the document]
dtolnay: This has not been addressed yet.

src/number.rs
dtolnay: This has not been addressed yet.

alexreg: See my comment in the main thread.

src/number.rs
dtolnay: People who use num-traits 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, or 0.9 may not appreciate it as much as you think. Look at how much people love the Serde 0.8 impls in the `url` crate.

src/value/de.rs
dtolnay: This has not been addressed yet.

src/value/ser.rs
```diff
@@ -247,23 +266,33 @@ impl serde::Serializer for Serializer {
     }
 }
 
+/// Not public API. Should be pub(crate).
```
dtolnay: I don't think this and the following few similar comments are meaningful because these types are effectively already `pub(crate)`. They are not accessible from outside of serde_json. Please remove.

src/value/ser.rs
```diff
@@ -417,3 +643,175 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
+
+    fn serialize_none(self) -> Result<Self::Ok, Self::Error> {
+        Err(invalid_number())
+    }
+
+    fn serialize_some<T: ?Sized>(self, _value: &T) -> Result<Self::Ok, Self::Error>
+    where
+        T: Serialize,
+    {
+        Err(key_must_be_a_string())
```
dtolnay: Is this intended to be `invalid_number()` like the other serializer methods here?

src/value/ser.rs
```diff
@@ -387,25 +598,40 @@ impl serde::ser::SerializeStruct for SerializeMap {
...
+        }
     }
 }
 
 impl serde::ser::SerializeStructVariant for SerializeStructVariant {
     type Ok = Value;
     type Error = Error;
 
-    fn serialize_field<T: ?Sized>(&mut self, key: &'static str, value: &T) -> Result<(), Error>
+    fn serialize_field<T: ?Sized>(&mut self, key: &'static str, value: &T) -> Result<(), Self::Error>
```
dtolnay: I would prefer not to make this change in this PR. The existing code is inconsistent but let's leave it to a later PR to pick one usage and standardize everywhere.

src/value/ser.rs
```diff
@@ -417,3 +643,175 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
+    }
+
+    fn serialize_char(self, _v: char) -> Result<Self::Ok, Self::Error> {
+        Err(invalid_number())
+    }
+
+    fn serialize_str(self, value: &str) -> Result<Self::Ok, Self::Error> {
+        let n = Number::from_string_unchecked(value.to_owned());
+        *self.0 = Some(Value::Number(n));
+        Ok(())
```
dtolnay: I think this would be clearer if we use the `Self::Ok` return value to hand back the resulting Number. The caller can take care of placing that number where they want it.

```rust
Ok(Number::from_string...)
```

So-called "out parameters" like `&'a mut Option<Value>` are not idiomatic in Rust when we can help it.

src/value/ser.rs
```diff
@@ -417,3 +643,175 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
+    fn serialize_f64(self, _v: f64) -> Result<Self::Ok, Self::Error> {
+        Err(invalid_number())
+    }
+
+    fn serialize_char(self, _v: char) -> Result<Self::Ok, Self::Error> {
+        Err(invalid_number())
+    }
+
+    fn serialize_str(self, value: &str) -> Result<Self::Ok, Self::Error> {
+        let n = Number::from_string_unchecked(value.to_owned());
```
dtolnay: It looks like this `from_string_unchecked` method was copied from the specialization-based implementation. But in the specialization case it was not necessary to check the string because we could know it came from a number. Here that is not the case and checking would be sensible.

src/value/ser.rs
```diff
@@ -215,14 +222,13 @@ impl serde::Serializer for Serializer {
     }
 
     fn serialize_map(self, _len: Option<usize>) -> Result<Self::SerializeMap, Error> {
-        Ok(
-            SerializeMap {
-                map: Map::new(),
-                next_key: None,
-            },
-        )
+        Ok(SerializeMap::Map {
```
dtolnay: Please revert unnecessary whitespace changes like this one. We can run the new rustfmt after the PR lands.

src/value/ser.rs
```diff
@@ -348,34 +377,216 @@ impl serde::ser::SerializeMap for SerializeMap {
...
-            _ => return Err(Error::syntax(ErrorCode::KeyMustBeAString, 0, 0)),
-        };
-        Ok(())
+        match *self {
+            SerializeMap::Map { ref mut next_key, .. } => {
+                *next_key = Some(try!(key.serialize(MapKeySerializer)));
+                Ok(())
+            },
+            #[cfg(feature = "arbitrary_precision")]
+            SerializeMap::Number { .. } => panic!(),
```
dtolnay: Is this reachable? A malicious input should not be able to trigger a panic.

alexreg: No, it's not reachable AFAIK. I should probably change to `unreachable!()`.

src/value/ser.rs
alexreg: Hmm, not sure how that happened!

src/de.rs
```diff
@@ -281,6 +294,35 @@ impl<'de, R: Read<'de>> Deserializer<R> {
...
+    ) -> Result<Number>
+    {
+        self.parse_integer(pos)
+    }
+
+    #[cfg(feature = "arbitrary_precision")]
+    fn parse_any_number(
+        &mut self,
+        pos: bool,
+        prim: bool
```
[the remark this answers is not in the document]
alexreg: Agreed. Since the changes in the code during the review, this makes more sense now.

src/value/de.rs
alexreg: Done.

src/value/ser.rs
alexreg: Right. I can get rid of the `doc(hidden)` attributes too...

src/value/ser.rs
alexreg: Okay, sure. I'll leave that to someone else. ;-)

src/value/ser.rs
alexreg: Yes, I *think* so. Not 100% sure, since I can't remember my motivation for ever writing that... or if it was sloppy copy & paste coding, perhaps.

tests/test.rs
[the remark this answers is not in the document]
alexreg: Yep, will add these. A few integers and floats that are too big for u64/i64/f64 should be good enough, I think.

src/value/ser.rs
alexreg: Yep. Not sure why I did it this way originally... possibly my knowledge of Rust/Serde wasn't very good when I first wrote this!

alexreg: Note: I believe I still need to use an "out value" internally for `impl serde::ser::SerializeStruct for SerializeMap`.

src/number.rs
alexreg: Okay, fair point. But we'll still have the `num-traits` dependency, as before. Its usage just won't be exposed. Are you okay with that?

src/value/ser.rs
alexreg: I think it would also make sense to check in the `NumberFromString` `Visitor` in the `number` module?

Anyway, how do you recommend doing this check? I think we should add a `FromStr` implementation for `Number` to be honest, which does the checking... perhaps using the `scan_integer` function? Although this would necessitate the construction of a `Deserializer` value. If that doesn't seem to "heavy" an approach for you, I can give it a go.

alexreg: Actually, it seems the use of `from_string_unchecked` is okay here. Wouldn't `test_parse_number_errors` fail otherwise?

dtolnay:
Please check on the failing builds. Looks like the non arbitrary-precision code does not compile.

src/value/ser.rs
dtolnay: Please revert this and similar unnecessary formatting changes.

src/value/ser.rs
```diff
@@ -348,72 +371,441 @@ impl serde::ser::SerializeMap for SerializeMap {
...
+    fn serialize_f32(self, _value: f32) -> Result<Self::Ok, Self::Error> {
+        Err(key_must_be_a_string())
+    }
+
+    fn serialize_f64(self, _value: f64) -> Result<Self::Ok, Self::Error> {
+        Err(key_must_be_a_string())
+    }
+
+    fn serialize_char(self, _value: char) -> Result<Self::Ok, Self::Error> {
+        Err(key_must_be_a_string())
```
dtolnay: I believe this used to be allowed. The following code works with the released serde_json but not with this PR.

```rust
println!("{}", json!(({
    let mut map = BTreeMap::new();
    map.insert('c', ());
    map
})));
```

src/value/ser.rs
dtolnay: Fixed.

src/value/ser.rs
alexreg: @dtolnay

src/value/ser.rs
alexreg: My bad, I thought this was in code I introduced for some reason. Reverted, and left the inconsistency as-is.

src/value/ser.rs
alexreg: Good catch. Fixed now.

alexreg:
The CI failure looks like it's nothing to do with my PR.

dtolnay:
It was an error message improvement in rustc and a new Clippy lint, fixed in c58b5201e17e155933c24f27d781a7aade824ad6 and 92c23fb828e38f857d9653766200557b95d1ab3c.

alexreg:
@dtolnay Okay. Could you kindly address my above question, by the way?

src/value/ser.rs
dtolnay: The point of checking would be to uphold the invariant that in safe code you cannot get a Number holding anything other than what we consider a JSON number. This allows other code in serde_json to assume that the string inside of a Number contains something that is a number.

```rust
#[macro_use]
extern crate serde_derive;

extern crate serde_json;

#[derive(Serialize)]
#[serde(rename = "$__serde_private_Number")]
struct S {
    #[serde(rename = "$__serde_private_number")]
    f: &'static str,
}

fn main() {
    println!("{:#?}", serde_json::to_value(&S { f: "serde" }).unwrap());
}
```

A `rename` attribute is the simplest way to illustrate the failure mode; a more realistic example would be [transcoding](https://serde.rs/transcode.html) from [RON](https://github.com/ron-rs/ron) to JSON in which malicious RON input fed by an attacker is able to trigger exploitable undefined behavior by violating assumptions in downstream code.

src/ser.rs
```diff
@@ -346,24 +361,20 @@ where
                     .end_array(&mut self.writer)
                     .map_err(Error::io)
             );
-            Ok(
```
dtolnay: Please revert these formatting changes. If rustfmt is formatting this code differently these days, we can run rustfmt separately from this PR.

src/value/ser.rs
```diff
@@ -348,51 +371,255 @@ impl serde::ser::SerializeMap for SerializeMap {
...
     }
 }
 
 impl serde::ser::SerializeStruct for SerializeMap {
     type Ok = Value;
     type Error = Error;
 
-    fn serialize_field<T: ?Sized>(&mut self, key: &'static str, value: &T) -> Result<(), Error>
+    fn serialize_field<T: ?Sized>(&mut self, key: &'static str, value: &T) -> Result<(), Self::Error>
```
dtolnay: Please revert anywhere that you changed `Error` to `Self::Error` or vice versa. We can leave it to a later PR to pick one usage and standardize everywhere.

src/value/ser.rs
```diff
@@ -404,16 +631,186 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
     where
         T: Serialize,
     {
-        self.map
-            .insert(String::from(key), try!(to_value(&value)));
+        self.map.insert(String::from(key), try!(to_value(&value)));
```
dtolnay: Please revert this formatting change.

src/value/ser.rs
alexreg: Yeah... I'm having a hard time spotting where I did this, so I think you'll just have to continue pointing them out, I'm afraid.

tests/test.rs
```diff
@@ -568,6 +569,19 @@ fn test_write_newtype_struct() {
     test_encode_ok(&[(outer, r#"{"outer":{"inner":123}}"#)]);
 }
 
+#[test]
+fn test_deserialize_number_to_untagged_enum() {
+    use serde_json::Number;
```
dtolnay: This is already imported on line 42.

src/value/ser.rs
```diff
@@ -348,51 +371,255 @@ impl serde::ser::SerializeMap for SerializeMap {
...
+                } else {
+                    Err(invalid_number())
+                }
+            },
+        }
     }
 
-    fn end(self) -> Result<Value, Error> {
-        serde::ser::SerializeMap::end(self)
+    fn end(self) -> Result<Self::Ok, Self::Error> {
```
dtolnay: Please revert the change in this method signature.

src/value/ser.rs
```diff
@@ -404,16 +631,186 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
         T: Serialize,
     {
-        self.map
-            .insert(String::from(key), try!(to_value(&value)));
+        self.map.insert(String::from(key), try!(to_value(&value)));
         Ok(())
     }
 
-    fn end(self) -> Result<Value, Error> {
+    fn end(self) -> Result<Self::Ok, Self::Error> {
```
dtolnay: Please revert this change.

tests/test.rs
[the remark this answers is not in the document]
dtolnay: This still needs to be addressed.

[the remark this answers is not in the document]
alexreg: It essentially has been: there is a `test_parse_arbitrary_precision_number` fn in `tests`, it just has code commented out for the time being.

src/value/ser.rs
alexreg: @dtolnay What were your thoughts on my above comment by the way?

> Anyway, how do you recommend doing this check? I think we should add a FromStr implementation for Number to be honest, which does the checking... perhaps using the scan_integer function? Although this would necessitate the construction of a Deserializer value. If that doesn't seem to "heavy" an approach for you, I can give it a go.

I could even make it simpler and just do `value::from_str` and extract the `Number` from that... but again, is that too heavy? Would appreciate your thoughts here.

dtolnay: I would not consider it heavy to construct a Deserializer. Implementing FromStr for Number in terms of scan seems like it would work.

alexreg: Okay, I'll give that a go, ta.

src/ser.rs
alexreg: Done.

src/value/ser.rs
alexreg: Done.

src/value/ser.rs
alexreg: Done.

tests/test.rs
alexreg: Fixed.

src/value/ser.rs
alexreg: Done.

src/value/ser.rs
alexreg: Done.

src/value/ser.rs
alexreg: Done.

alexreg:
@dtolnay Okay, I believe all outstanding issues are in order now! I think we're very close to being able to merge. Please tell me I'm right. 😉

src/number.rs
```diff
@@ -292,90 +381,355 @@ impl<'de> Deserialize<'de> for Number {
...
+        bytes byte_buf map struct option unit newtype_struct
+        ignored_any unit_struct tuple_struct tuple enum identifier
+    }
+}
+
+impl FromStr for Number {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        super::de::Deserializer::from_str(s).parse_any_signed_number().map(|n| n.into())
```
dtolnay: Please make this act as much as possible like the FromStr for primitive types. For example the leading whitespace behavior should be the same. Compare:

```rust
extern crate serde_json;

fn main() {
    println!("{}", " 0".parse::<serde_json::Number>().unwrap());
    println!("{}", " 0".parse::<u64>().unwrap_err());
}
```

src/number.rs
```diff
@@ -292,90 +381,355 @@ impl<'de> Deserialize<'de> for Number {
...
+        struct Visitor;
+
+        impl<'de> de::Visitor<'de> for Visitor {
+            type Value = NumberFromString;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("string containing a number")
+            }
+
+            fn visit_string<E>(self, s: String) -> Result<NumberFromString, E>
```
dtolnay: According to the [`Visitor::visit_string`](https://docs.serde.rs/serde/de/trait.Visitor.html#method.visit_string) documentation:

> It is never correct to implement `visit_string` without implementing `visit_str`.

In this case I think you want just `visit_str` because the implementation does not benefit from taking ownership of the String data.

alexreg: Fair point.

src/number.rs
alexreg: Sure, I'll give it a go. Unfortunately, the `ui` tests are failing again now...

alexreg:
@dtolnay Question: should we implement the experimental `TryFrom` trait for floats to `Number`s?

src/number.rs
alexreg: Fixed. As for the primitive types, leading and trailing whitespace now cause errors.

alexreg:
@dtolnay How are we looking now?

src/ser.rs
```diff
@@ -1304,6 +1554,14 @@ pub trait Formatter {
         dtoa::write(writer, value).map(|_| ())
     }
 
+    /// Writes a number that has already been rendered to a string.
+    #[inline]
+    fn write_number_str<W: ?Sized>(&mut self, writer: &mut W, value: &str) -> Result<()>
```
dtolnay: The returned Result type should be io::Result instead of serde_json::Result to be consistent with the rest of the Formatter methods.

src/ser.rs
```diff
@@ -1148,6 +1236,168 @@ where
...
+    }
+
+    fn serialize_none(self) -> Result<Self::Ok> {
+        Err(invalid_number())
+    }
+
+    fn serialize_some<T: ?Sized>(self, _value: &T) -> Result<Self::Ok>
+        where T: Serialize
+    {
+        Err(key_must_be_a_string())
```
dtolnay: I think we saw this somewhere else before, the error here is inconsistent with the `invalid_number()` in all the other methods.

dtolnay:
> Question: should we implement the experimental `TryFrom` trait for floats to `Number`s?

I would not do this yet until someone has a compelling real use case that benefits from TryFrom.

alexreg:
@dtolnay Fair enough. How are we doing now; ready to merge? :-)

dtolnay:
Fix the two review comments first. :-)

src/ser.rs
alexreg: Yep, copy & paste coding. :-P

alexreg:
> Fix the two review comments first. :-)

Serves me right for replying to your comment before I open up the issue. (GitHub notifies me for normal replies, but not review comments.)

dtolnay:
You may need to check the box for "Pull Request reviews" on https://github.com/settings/notifications.

alexreg:
@dtolnay Yep. I think I turned it off because I was getting spam from reviews on another project. Seems it can't be enabled on a per-PR/repo basis. Anyway, I think we're coming to the end of this review now. Let me know if there's anything else to do before we can get this merged, and I'll try to fix it all tonight!

alexreg:
No problem, @dtolnay. Thanks for your patience, and guiding me through. A review for such a large PR was never going to be simple.

PR #421, opened 2018-03-11
Add `Value::take` method

dmizuk:
This is equivalent to `Option::take` method.


dtolnay:
I released serde_json 1.0.11 with this method.

Issue #420, opened 2018-03-10
Rename 'pos' to 'positive'


dtolnay:
We use `pos` in a few places related to number parsing to mean `positive`. The shortened name is ambiguous with `position` so let's spend the extra characters to clarify it.

oli-obk:
`pos` has been fixed, there are now just things like `exp_pos` left.

PR #425, opened 2018-03-20
Replace relevant instance of "pos" with "positive"

mtn:
Addresses issue #420 

This library has been super helpful to me! Thank you for the hard work 🎈 

I grepped for instances of `pos` and replaced the ones which seemed relevant. I only replaced full abbreviations, leaving things like `exp_pos` alone. I also ran the tests locally, though I didn't check that they passed for every single toolchain.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -281,13 +281,13 @@ impl<'de, R: Read<'de>> Deserializer<R> {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


oli-obk:
Thanks! great cleanup.

> leaving things like exp_pos alone

I'll add it to the issue so we don't forget about it.

PR #439, opened 2018-05-01
Implement order-independent equality

### LINKED ISSUE #438: `preserve_order` changes semantics of equality on `Value`
Without `preserve_order` it is safe to compare two `Value`s for equality. However, with the feature enabled, two maps with the same keys may no longer compare equal:

https://contain-rs.github.io/linked-hash-map/src/linked_hash_map/lib.rs.html#760-764

This goes against the idea of cargo features being "additive", and can cause some very tricky to locate bugs...

IMO, the `Value` type should have consistent equality semantics, and you could add a wrapper type for retaining the original behaviour.

Diggsey:
Fixes #438

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -230,7 +230,11 @@ impl Clone for Map<String, Value> {
 impl PartialEq for Map<String, Value> {
     #[inline]
     fn eq(&self, other: &Self) -> bool {
-        self.map.eq(&other.map)
+        if self.len() != other.len() {
+            return false;
+        }
+
+        self.iter().all(|(key, value)| other.get(key).map_or(false, |v| *value == *v))
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/map.rs
```diff
@@ -230,7 +230,11 @@ impl Clone for Map<String, Value> {
 impl PartialEq for Map<String, Value> {
     #[inline]
     fn eq(&self, other: &Self) -> bool {
-        self.map.eq(&other.map)
```
dtolnay: Please keep the existing implementation if we are building without preserve_order. I believe BTreeMap will be able to determine equality significantly faster.

Diggsey: Fixed

dtolnay:
I released this fix in serde_json 1.0.17.

PR #451, opened 2018-06-06
Use IndexMap instead of LinkedHashMap

lnicola:
This was surprisingly easy :confused:.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ serde_derive = "1.0"
 [features]
 default = []
 
-# Use IndexMap rather than BTreeMap as the map type of serde_json::Value.
+# Use a different representation for the map type of serde_json::Value.
 # This allows data to be read into a Value and written back to a JSON string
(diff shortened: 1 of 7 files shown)


lnicola:
Is there a benchmark suite or something like that? I wonder if we could drop the other implementation.

lnicola:
Oh, `indexmap` requires 1.18.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ appveyor = { repository = "serde-rs/json" }
 
 [dependencies]
 serde = "1.0.60"
-linked-hash-map = { version = "0.5", optional = true }
+indexmap = { version = "1.0", optional = true }
```
dtolnay: Please change the Travis configuration to add a build at indexmap's minimum supported compiler version, and remove any `preserve_order` builds below that compiler version.

lnicola: Done

PR #453, opened 2018-06-15
Allow char to be a MapKey

fdionisi:
I think `char`s should be allowed to be `MapKey` or would that be in some way harmful?

I'm new to Rust and this repository seems really complex to me 😬.
Please let me know if and where I have to add/update tests – I'll be more than happy to do that!

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1140,8 +1140,8 @@ where
         Err(key_must_be_a_string())
     }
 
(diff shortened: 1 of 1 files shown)


dtolnay:
Thank you. I added some tests in 9316c316257aed02e41b7d5a2cee29e620618e73 and published this as 1.0.21.

Issue #355, opened 2017-09-03
Add a RawValue type


alteous:
It would be helpful to have a type similar to Go's [`json.RawMessage`](https://golang.org/pkg/encoding/json/#RawMessage) that is not tokenized during deserialization, but rather its raw contents stored as a `Vec<u8>` or `&'de [u8]`.

The following pseudo-code demonstrates the idea.
```rust
#[derive(Deserialize)]
struct Struct {
    /// Deserialized normally.
    core_data: Vec<i32>,

    /// Contents of `user_data` are copied / borrowed directly from the input
    /// with no modification.
    ///
    /// `RawValue<'static>` is akin to `Vec<u8>`.
    /// `RawValue<'a>` is akin to `&'a [u8]`.
    user_data: serde_json::RawValue<'static>,
}

fn main() {
    let json = r#"
    {
        "core_data": [1, 2, 3],
        "user_data": { "foo": {}, "bar": 123, "baz": "abc" }
    }
    "#;
    
    let s: Struct = serde_json::from_bytes(&json).unwrap();
    println!("{}", s.user_data); // "{ \"foo\": {}, \"bar\": 123, \"baz\": \"abc\" }"
}
```

The main advantage of this is would be to have 'lazily-deserialized' values.

shepmaster:
@dtolnay mentions that the work happening in #416 might lay some groundwork for such a change to take place.

alexreg:
Yeah, this sounds useful. Ideally we could get some of the building blocks for this sort of thing (custom 'built-in' types) integrated into serde itself, in time.

srijs:
I need this for a project I'm working on, and would be happy to give it shot following the same approach as #416. Is this something that you'd accept a PR for?

dtolnay:
Yes I would like a PR for this. When you say you need this feature, is it because deserializing to serde_json::Value is too slow in your project and you are expecting this to have better performance, or some other reason?

dtolnay:
As a first step, please put together a short benchmark that you trust as reasonably representative of the shuffling data back and forth piece of your application. Ideally it would depend on serde / serde_derive / serde_json but nothing else -- so no postgres or http stuff. Focus on the data structures so that they have the same ratio and pattern of raw vs meaningful content as your real use case. Write it using serde_json::Value for now, and then the goal will be to swap in a raw value once it's implemented and observe a speedup.

srijs:
[...]

Issue #323, opened 2017-05-26
Experiment with a single-allocation Value implementation


dtolnay:
The sajson library has an interesting [single allocation mode](https://github.com/chadaustin/sajson#single) where it pre-allocates a buffer of size one word per byte in the input document. Justified in [this blog post](https://chadaustin.me/2013/01/sajson-why-the-parse-tree-is-big-enough/). Then it lays out data structures in that buffer without needing a single bounds check along the way.

Once #318 has been implemented, we can experiment with something similar. It should be possible to expose a Deserialize implementation that works like Value but performs just one allocation up front and no bounds checks.

PR #480, opened 2018-09-14
Implement RawValue type

srijs:
Implements the `RawValue` type described in #355. On [this benchmark](https://gist.github.com/srijs/7fe21b43c2cf5d2ceeb593ae4591c6f5), using the `RawValue` type instead of deserializing/serializing via `Value` results in a 3-4x speed-up.

```
test deserialize_and_serialize_raw_value ... bench:       1,759 ns/iter (+/- 354)
test deserialize_and_serialize_value     ... bench:       6,091 ns/iter (+/- 1,776)
```

I did not adhere fully to the way that the existing arbitrary precision feature is implemented, because I found it easier to use newtypes rather than structs to pass the raw data back and forth. If you want, I'm happy to either change it so that it uses the same approach as the arbitrary precision feature, or to change the arbitrary precision feature to use this (imo) simpler approach.

In the documentation for `RawValue` I mentioned two important caveats which apply for this implementation:

1. When serializing, a `RawValue` will retain its original formatting and will not be minified or pretty-printed.
2. When deserializing, `RawValue` can not be used with the `#[serde(flatten)]` attribute, as it relies on the original input buffer.

What you described in #323 might be a cleaner solution to these problems, but it also sounds more complex and resource hungry (using 4x-8x the amount of memory).

Let me know what you think!

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -44,7 +44,7 @@ use serde_bytes::{ByteBuf, Bytes};
 
 use serde_json::{
     from_reader, from_slice, from_str, from_value, to_string, to_string_pretty, to_value, to_vec,
-    to_writer, Deserializer, Number, Value,
+    to_writer, Deserializer, Number, RawValue, Value,
 };
 
 macro_rules! treemap {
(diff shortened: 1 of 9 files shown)


alteous:
Awesome! Thank you. :+1:

tests/test.rs
```diff
@@ -2039,3 +2039,52 @@ fn test_integer128() {
...
+        #[serde(borrow)]
+        b: RawValue<'a>,
+        c: i8,
+    };
+
+    let wrapper_from_str =
+        serde_json::from_str::<Wrapper>(r#"{"a": 1, "b": {"foo": 2}, "c": 3}"#).unwrap();
+    assert_eq!(r#"{"foo": 2}"#, wrapper_from_str.b.as_ref());
+
+    let wrapper_from_reader = serde_json::from_reader::<_, Wrapper<'static>>(
```
dtolnay: This is a compiler bug, this line shouldn't compile. I filed https://github.com/rust-lang/rust/issues/54302 to follow up. Still thinking about what that means for how we want to expose RawValue.

srijs: Hah, of course, not sure what I thought this would do!

Does that mean we want to have an owned and a borrowed variant of `RawValue`?

dtolnay: I believe so. I opened #485 to experiment with that design. Let me know what you think.

dtolnay:
> I did not adhere fully to the way that the existing arbitrary precision feature is implemented, because I found it easier to use newtypes rather than structs to pass the raw data back and forth. If you want, I'm happy to either change it so that it uses the same approach as the arbitrary precision feature, or to change the arbitrary precision feature to use this (imo) simpler approach.

The way `Number` does it is more robust than newtypes when mixing with data formats that are not JSON. See 7e373b6125ee6f19bb9a9538c51c92bf51be1f90.

srijs:
Ah, I was looking for an explanation like that, might be worth adding a code comment to make this clear to anyone coming across this in the future.

As I said on the other PR (#485), do let me know if I can help move this forward in any way!

PR #485, opened 2018-09-20
Implement RawValue type (alternative)

dtolnay:
This PR builds on #480 but eliminates the `Cow` in favor of distinct `&RawValue` and `Box<RawValue>` options.

Documentation of `RawValue`:

---

Reference to a range of bytes encompassing a single valid JSON value in the input data.

A `RawValue` can be used to defer parsing parts of a payload until later, or to avoid parsing it at all in the case that part of the payload just needs to be transferred verbatim into a different output object.

When serializing, a value of this type will retain its original formatting and will not be minified or pretty-printed.

### Example

```rust
#[macro_use]
extern crate serde_derive;
extern crate serde_json;

use serde_json::{Result, value::RawValue};

#[derive(Deserialize)]
struct Input<'a> {
    code: u32,
    #[serde(borrow)]
    payload: &'a RawValue,
}

#[derive(Serialize)]
struct Output<'a> {
    info: (u32, &'a RawValue),
}

// Efficiently rearrange JSON input containing separate "code" and "payload"
// keys into a single "info" key holding an array of code and payload.
//
// This could be done equivalently using serde_json::Value as the type for
// payload, but &RawValue will perform netter because it does not require
// memory allocation. The correct range of bytes is borrowed from the input
// data and pasted verbatim into the output.
fn rearrange(input: &str) -> Result<String> {
    let input: Input = serde_json::from_str(input)?;

    let output = Output {
        info: (input.code, input.payload),
    };

    serde_json::to_string(&output)
}

fn main() -> Result<()> {
    let out = rearrange(r#" {"code": 200, "payload": {}} "#)?;

    assert_eq!(out, r#"{"info":[200,{}]}"#);

    Ok(())
}
```

### Ownership

The typical usage of `RawValue` will be in the borrowed form:

```rust
#[derive(Deserialize)]
struct SomeStruct<'a> {
    #[serde(borrow)]
    raw_value: &'a RawValue,
}
```

The borrowed form is suitable when deserializing through [`serde_json::from_str`] and [`serde_json::from_slice`] which support borrowing from the input data without memory allocation.

When deserializing through [`serde_json::from_reader`] you will need to use the boxed form of `RawValue` instead. This is almost as efficient but involves buffering the raw value from the I/O stream into memory.

[`serde_json::from_str`]: https://docs.serde.rs/serde_json/fn.from_str.html
[`serde_json::from_slice`]: https://docs.serde.rs/serde_json/fn.from_slice.html
[`serde_json::from_reader`]: https://docs.serde.rs/serde_json/fn.from_reader.html

```rust
#[derive(Deserialize)]
struct SomeStruct {
    raw_value: Box<RawValue>,
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,12 @@ compiletest_rs = "0.3"
 serde_bytes = "0.10"
 serde_derive = "1.0"
 
+[package.metadata.docs.rs]
+features = ["raw_value"]
+
+[package.metadata.playground]
+features = ["raw_value"]
+
 
 ### FEATURES #################################################################
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,3 +41,6 @@ preserve_order = ["indexmap"]
 # allows JSON numbers of arbitrary size/precision to be read into a Number and
 # written back to a JSON string without loss of precision.
 arbitrary_precision = []
+
+# Provide a RawValue type that can hold unprocessed JSON during deserialization.
+raw_value = []
--- a/src/de.rs
+++ b/src/de.rs
@@ -8,7 +8,6 @@
 
 //! Deserialize JSON data to a Rust data structure.
 
-use std::borrow::Cow;
 use std::io;
 use std::marker::PhantomData;
 use std::result;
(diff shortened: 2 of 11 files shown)


srijs:
I like where this is going! Anything I can pick up or help with?

PR #498, opened 2018-10-14
Add i128 support

### LINKED ISSUE #487: impl as_u128/as_i128 for Number
this should involve implementing `visit_u128`/`visit_i128` for the Number type if i'm not wrong?

macisamuele:
This fixes #487.

This PR adds support to 128-bits integers.
To avoid the introduction of a lot of boilerplate and code duplication I've modified the `number::N` to hold 128-bits integer.

I've not verified the whole library yet to ensure that all the integers flows are covered.
I opened this to have an initial feedback about the process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,15 +11,9 @@ matrix:
         - cargo test --features arbitrary_precision
         - cargo test --features raw_value
 
-    - rust: 1.15.0
-      script:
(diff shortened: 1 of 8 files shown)


rslife:
Hello @dtolnay.

Now that almost a year has passed, would you be open to re-considering raising the minimum Rust version required?

Related issue: pyfisch/cbor#140

PR #508, opened 2019-01-11
Support parsing of JSON with arbitrary depth

### LINKED ISSUE #334: Allow increasing recursion limit

RReverser:
This introduces a feature flag that would use [`stacker`](https://github.com/alexcrichton/stacker) to prevent stack overflow while parsing a deeply nested JSON.

Benchmarks show no change with default feature flags (despite extracting a custom helper for recursion), and a relatively small cost for stack probes when nesting into arrays / objects with a feature enabled.

This doesn't attempt to fix all the other operations on the parsed result, and so `PartialEq` / `ParialOrd` / `Display` / `Debug` / `Drop` / ... can still stack overflow after certain depth as they normally would in Rust.

Fixes #334.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
         - cargo test --features preserve_order
         - cargo test --features arbitrary_precision
         - cargo test --features raw_value
+        - cargo test --features arbitrary_depth
 
     - rust: 1.15.0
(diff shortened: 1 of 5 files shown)


RReverser:
Closing in favour of #509.

PR #509, opened 2019-01-13
Unbounded depth

dtolnay:
Provides a method `disable_recursion_limit` to parse arbitrarily deep JSON structures without any consideration for overflowing the stack.

When using this method, the caller will want to provide some other way to protect against stack overflows, such as by wrapping the Deserializer in the dynamically growing stack adapter provided by the [serde_stacker] crate.

[serde_stacker]: https://github.com/dtolnay/serde-stacker

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -42,12 +42,23 @@ where
     ///   - Deserializer::from_bytes
     ///   - Deserializer::from_reader
     pub fn new(read: R) -> Self {
-        Deserializer {
-            read: read,
-            scratch: Vec::new(),
-            remaining_depth: 128,
-            #[cfg(feature = "unbounded_depth")]
-            disable_recursion_limit: false,
+        #[cfg(not(feature = "unbounded_depth"))]
+        {
+            Deserializer {
+                read: read,
+                scratch: Vec::new(),
+                remaining_depth: 128,
+            }
+        }
+
+        #[cfg(feature = "unbounded_depth")]
+        {
+            Deserializer {
+                read: read,
+                scratch: Vec::new(),
+                remaining_depth: 128,
+                disable_recursion_limit: false,
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/de.rs
```diff
@@ -40,10 +42,23 @@ where
...
+            }
+        }
+
+        #[cfg(feature = "unbounded_depth")]
+        {
+            Deserializer {
+                read: read,
+                scratch: Vec::new(),
+                remaining_depth: 128,
+                disable_recursion_limit: false,
```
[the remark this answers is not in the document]
dtolnay: The approach in your PR fails to parse on versions of Rust that we support, so we can't do it that way.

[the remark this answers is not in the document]
RReverser: Huh weird. I'd never expect it to be a new-ish feature.

src/de.rs
```diff
@@ -140,6 +155,54 @@ impl<'de, R: Read<'de>> Deserializer<R> {
...
+    ///     let mut stack = vec![value];
+    ///     while let Some(value) = stack.pop() {
+    ///         if let Value::Array(array) = value {
+    ///             stack.extend(array);
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    #[cfg(feature = "unbounded_depth")]
+    pub fn disable_recursion_limit(&mut self) {
```
RReverser: Since we anyway go for adding a dedicated method, I would probably accept `depth: Option<u8>` or `depth: Option<usize>` instead for flexibility, but at least being able to disable a hardcoded constant depth is already good enough.

dtolnay: I think disabling serde_json's limit is the only essential change. A different recursion limit could be imposed through an adapter outside of serde_json. I filed https://github.com/dtolnay/serde-stacker/issues/1 to include that in serde_stacker.

RReverser: It could, I just don't necessarily see a need for +1 dependency when there is already a field which could be exposed for users who merely want to change the limit.

RReverser: Another alternative would be removing magic constant altogether and leave any stack guards to the `serde-stacker`.

After all, the stack exhaustion issue is not specific to JSON in any way as you mentioned yourself, and I don't see why it has to be handled within this specific adapter.

RReverser: @dtolnay Just in case this got lost - how does this sound / what are your plans about this PR?

dtolnay:
Released in 1.0.38.

Issue #168, opened 2016-11-24
Allow comments in JSON


selaux:
Hi we are trying to use serde-json in a C++/Rust. The C++ library we use allows for comments in the JSON files. We would like to have the same in rust. Could that be a PR for this library? Or is that not possible because it's not part of the Spec? How would we go about writing our own parser without copying code from serde-json?

dtolnay:
I would strongly prefer not to support this because it is not part of the spec. [Hjson](https://github.com/hjson/hjson-rust) may be better for your use case. It should be able to parse "JSON with comments."

dtolnay:
Let me know if Hjson does not work for your use case and we can reconsider.

nicoburns:
I'd like to make a case for supporting `//` line comments only in `serde_json`:

* Rightly or wrongly, JSON is frequently used for shared config files in multi-language environments. And for this use case, allowing comments in the file is often desirable. Notably there isn't really a good alternative to JSON+comments for this use case. TOML doesn't have good cross-language support, and YAML is considerably more complex.
* The other parts of "lax JSON" specs such as JSON5 or HJSON are often *not* desirable in this situation. For example, if a config file is missing quotes around it's key, the correct solution is to modify the config file to include the quotes.
* The same is not true for comments, which it is often desriable to keep for documentation purposes.
* Adding support for `//` single-line comments is a very minor change to the parser, and is unlikely to meaningfully affect parsing performance.
* `serde_json` is much more actively developed than other projects such as `Hjson` (For example, at the time of writing, `Hjson` hasn't yet been updated to be compatible with serde 0.9. `serde_json` also has ). Has much more attention paid to it for things like performance and semantics around serializing enums, which it would be a shame to lose just for comment support.

TL;DR: I this use case is very common, has clear motivation, and has only a small cost.

oli-obk:
One big issue is that roundtripping won't work, because comments can't be deserialized in any useful manner.

> this use case is very common

That's subjective. I've never seen commented json.

> and has only a small cost.

If the cost is small enough (both in performance degradation and code complexity), I can live with it. Maybe it could be a `feature`.

erickt:
[...]

PR #514, opened 2019-01-24
[RFC] Add an option to allow C99-like comments.

bolinfest:
This would introduce a `serde_json::Options` type that could be
passed to `from_str`, `from_bytes, or `from_reader`. Initially,
the only option would be `allow_comments`, though I would also
like to add support for `allow_trailing_commas` going forward.

I read through https://github.com/serde-rs/json/issues/168 and
saw that this was rejected in the past, but I was curious if we
could revisit this issue. The suggestion on the issue was to use
[Hjson](https://github.com/hjson/hjson-rust), which I think is
"too loose" for most peoples' taste. Also it has [not been updated to
target serde 1.0.x yet](https://github.com/hjson/hjson-rust/issues/6).

Allowing C99-like comments and trailing commas results in an input
language that is still valid ECMAScript (it's just version 5.1 instead
of 3), which cannot be said for things like Hjson. Because it is
familiar to JavaScript developers, I have seen it used as an input
language in a number of domains. For example, all settings files in VS
Code appear to allow this: it even has a special language mode to
recognize this named "JSON with Comments (jsonc)."

It is fairly common for JSON parsers to allow options like this.
For example, here is the full set of options supported by Folly's
(C++ library) JSON parser:

https://github.com/facebook/folly/blob/74ea53886c3333eda4eaf457d538a678ceaa5add/folly/json.h#L56-L72

Google's Gson parser has a `setLenient()` option:

https://github.com/google/gson/blob/9d44cbc19a73b45971c4ecb33c8d34d673afa210/gson/src/main/java/com/google/gson/stream/JsonReader.java#L296-L324

There's also https://json5.org/, which tries to formalize the leniency.

If `serde_json` is not on board with this direction, I guess I should
fork this repo and create `serde_jsonc` or `serde_lenient_json`? I would
prefer not to do that, but if you want to keep `serde_json` so that it
only supports RFC 8259 or whatever it is, then I suppose forking is the
best way forward.

One final note: although https://github.com/serde-rs/json/issues/168
notes that comments were "explicitly excluded from JSON," it doesn't
mean it was the right choice. I've been arguing against it for years
now: http://bolinfest.com/essays/json.html.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -20,11 +20,26 @@ use number::NumberDeserializer;
 
 //////////////////////////////////////////////////////////////////////////////
 
+/// Options for a Deserializer.
+pub struct Options {
+    /// Whether to allow // and /* */ comments in the JSON.
+    pub allow_comments: bool,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            allow_comments: false,
+        }
+    }
+}
+
 /// A structure that deserializes JSON into Rust values.
 pub struct Deserializer<R> {
     read: R,
     scratch: Vec<u8>,
     remaining_depth: u8,
+    options: Options,
 }
 
 impl<'de, R> Deserializer<R>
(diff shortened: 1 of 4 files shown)


bolinfest:
Also, I know the actual code needs more comments and tests. For now, I just wanted to get enough code written to motivate the discussion. It's nice that it appears the support for this feature can be encapsulated within the `parse_whitespace()` function.

Issue #362, opened 2017-10-01
Support no_std


dtolnay:
We have been putting this off because nobody has needed it so far.

But in https://github.com/dtolnay/dtoa/issues/8:

> **\<pierre-rouanet>** I'm currently trying to use serde_json on a microcontroller

japaric:
FWIW, this morning I wrote a [serde-json-core](https://github.com/japaric/serde-json-core) crate that provides JSON (de)serialization on no_std. It's certainly not on par with the real serde-json as I don't plan to add any stuff that requires dynamic memory allocation to it, or add support for floats. I also made some adjustments to avoid having the binary size of my firmware increase by several KBs. Anyway, you can check the [documentation](https://japaric.github.io/serde-json-core/serde_json_core/) for details about what works and what doesn't work.

Also, *I* don't plan to send a PR to serde-json as I doubt all my changes are going to be welcomed, but anyone else is free to try.

cc @pierre-rouanet @perlindgren @MrBuddyCasino

dtolnay:
I added a link to serde-json-core in our readme and rustdoc, as well as https://serde.rs/no-std.html. Thanks!

dtolnay:
Reopening because there continues not to be a no_std JSON crate that can support types from `alloc` (String, Vec). https://github.com/japaric/serde-json-core/issues/4
I would be willing to consider a PR that adds a `std` feature on by default, and exposes the same or a subset of the current API of serde_json by using the `alloc` crate in no_std mode.

jonimake:
> Reopening because there continues not to be a no_std JSON crate that can support types from `alloc` (String, Vec). [japaric/serde-json-core#4](https://github.com/japaric/serde-json-core/issues/4)
> I would be willing to consider a PR that adds a `std` feature on by default, and exposes the same or a subset of the current API of serde_json by using the `alloc` crate in no_std mode.

I took a glance at how possible that would be, and I stumbled upon the fact that std::error isn't in core. What would you suggest to return if not std::error types in no_std situation?

geofft:
[...]

PR #516, opened 2019-02-01
No_std support

r-marques:
_Here be dragons_

This is an attempt at resolving #362.

---

This pr is far from complete. The goal is to start a discussion on what the effort would be to correctly add no_std support to serde_json.

What was done so far:
- Added a std feature that is enabled by default
- Added a no_std feature so that we can enable specific dependencie features as required (e.g. alloc in serde)
- Added no_std support for the deserializer

Build with no_std:
```bash
cargo build --no-default-features --features no_std

# building for a different target
rustup target add thumbv7m-none-eabi
cargo build --no-default-features --features no_std --target thumbv7m-none-eabi
```

This was tested in another project were we use serde_json for json deserialization. We were able successfully compile with no_std support using the changes in this pr.

---
**Problems:**

The import logic can get messy:
- Nested imports makes it cleaner but they were only added in rust version 1.25
- We can probably try to reuse the import logic performed by serde. Although it's weird to depend on how another crate does imports

Requires nightly:
- Both serde `unstable` and `alloc` features require nightly

Cannot have dev-dependencies:
- There seems to be a [problem in cargo](https://github.com/rust-lang/cargo/issues/4866) where the dependencies and dev-dependecies features are unified. With the dev-dependecies I couldn't find a way to disable serde default features
- This also means that we cannot run the tests

Error handling becomes a bit more difficult, specially parts of the code that make heavy use of std::io::Error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,15 +16,15 @@ travis-ci = { repository = "serde-rs/json" }
 appveyor = { repository = "serde-rs/json" }
 
 [dependencies]
-serde = "1.0.60"
-indexmap = { version = "1.0", optional = true }
-itoa = "0.4.3"
-ryu = "0.2"
+serde = { version = "1.0.85", default-features = false, features = ["alloc"] }
+#indexmap = { version = "1.0", optional = true }
+itoa = { version = "0.4.3", optional = true, default-features = false }
+ryu = { version = "0.2", optional = true, default-features = false }
 
 [dev-dependencies]
-compiletest_rs = { version = "0.3", features = ["stable"] }
-serde_bytes = "0.10"
-serde_derive = "1.0"
+# compiletest_rs = { version = "0.3", features = ["stable"] }
+# serde_bytes = "0.10"
+# serde_derive = "1.0"
 
 [package.metadata.docs.rs]
 features = ["raw_value"]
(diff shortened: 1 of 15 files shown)


dtolnay:
Thanks, looks like this is in the right direction.

r-marques:
Just an update.

I added no_std support for the Serializer. For now I skipped the PrettyFormatter and the the raw_value and arbitrary_precision features.

My main concern now is to figure out a way to handle dev-dependencies (since it overrides the features) so that I can run the tests.

When you have the time let me know if you have any suggestions for the code so far.

gbip:
Is there anything that could be done to help you finish this PR ? I have huge interest in this feature. :smile:

r-marques:
Right now the biggest problem is the fact that cargo unifies the `dependecies` and `dev-dependecies` which makes it difficult to run the tests. I don't really have a solution for that.

I would also need some input from @dtolnay to figure out how to proceed.

This merge request initially started from a previous project I was working on that was using serde_json. The project was using my [fork of serde_json](https://github.com/r-marques/json/tree/no_std_support) running on a cortex-m4f. In the meantime you can point cargo to my branch if you want to try it out.

gbip:
Is your project running on a cortex-m4f open source ?
I could use some inspiration !

r-marques:
Unfortunately the project although in a running state was abandoned and is currently closed source. I don't know at this point if the company will ever open source it.

But the rust embedded working group has great resources and the [discovery book](https://rust-embedded.github.io/discovery/index.html) is actually using the cortex-m4f

Issue #524, opened 2019-03-15
StreamDeserializer errors on cut-off decimal numbers


17dec:
Simple example:
```rust
fn main() {
    dbg!(serde_json::Deserializer::from_slice(b"[1.").into_iter::<Vec<f64>>().next());
}
```
I'd expect that to return None or an EOF error - the array is not finished yet and the number is valid if it's followed by more digits. But instead it returns an early parse error:
```rust
serde_json::Deserializer::from_slice(b"[1.").into_iter::<Vec<f64>>().next() = Some(
    Err(
        Error("invalid number", line: 1, column: 3)
    )
)
```

(What I'm really doing is using a StreamDeserializer with a little bit of custom parsing to parse a large array of values, so my problem would also be solved with #404 )

dtolnay:
Thanks, I would accept a PR to return an EOF error in this case.

PR #525, opened 2019-03-19
Return EOF on truncated numbers

17dec:
My attempt at fixing #524. First commit fixes that issue, the second commit also fixes a similar issue with truncated negative numbers ("-") and truncated exponents ("1e" and "1e+").

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -380,7 +380,14 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     }
 
     fn parse_integer(&mut self, positive: bool) -> Result<ParserNumber> {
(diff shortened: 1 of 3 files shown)


dtolnay:
Please add the email you used for these commits to https://github.com/settings/emails so that GitHub can attribute the commits to you.

17dec:
> Please add the email you used for these commits to https://github.com/settings/emails so that GitHub can attribute the commits to you.

Is that... a requirement to contribution? Because I'd prefer not to have all my git commits over the past 10 years linked to an account I otherwise never use.

dtolnay:
Thanks for checking -- it is not a requirement to contribute.

Issue #404, opened 2018-01-12
Provide a way for StreamDeserializer to stream an array of values


dtolnay:
It seems like a common situation to have a [large](https://github.com/serde-rs/json/issues/403) top-level array in a JSON file.

```rust
[
  {...},
  {...},
  {...},
  ...
]
```

StreamDeserializer should expose a way to access the elements of such an array one at a time without holding the whole thing in memory.

rofrol:
Example https://github.com/serde-rs/serde/issues/1107#issuecomment-348548681

davidOSUL:
Posting in case it helps someone else -- this isn't a solution to the issue to the StreamDeserializer issue, but one workaround to be able to access elements of a large json array one at a time is to use something like a [sync_channel](https://doc.rust-lang.org/std/sync/mpsc/fn.sync_channel.html) with a bound of zero, and have the sender do what the visitor in [this comment](https://github.com/serde-rs/serde/issues/1107#issuecomment-408729053) is doing. Here's a minimal example of that, I think you could probably also make the iterator and visitor generic on any type that is Deserializable as well: 

```rust
use serde::de::{Deserializer, SeqAccess, Visitor};
use serde::Deserialize;
use serde_json;
use std::fs::File;
use std::io::{BufReader, Write};
use std::path::PathBuf;
use std::sync::mpsc::{sync_channel, Receiver, SyncSender};
use std::{fmt, thread};

type DeserializeResult = Result<MyJson, String>;

#[derive(Deserialize, Debug)]
struct MyJson {
    val1: String,
    val2: Vec<i32>,
}

struct MyJsonIterator {
    receiver: Receiver<DeserializeResult>,
}

struct MyJsonVisitor {
    sender: SyncSender<DeserializeResult>,
}

impl Iterator for MyJsonIterator {
    type Item = DeserializeResult;

    fn next(&mut self) -> Option<Self::Item> {
        self.receiver.recv().ok() //ok() because a RecvError implies we are done
    }
}

impl MyJsonIterator {
    pub fn new(path: PathBuf) -> Self {
        let (sender, receiver) = sync_channel::<DeserializeResult>(0);

        thread::spawn(move || {
            let reader = BufReader::new(File::open(path).unwrap()); //in real scenario may want to send error, instead of unwrapping
            let mut deserializer = serde_json::Deserializer::from_reader(reader);
            if let Err(e) = deserializer.deserialize_seq(MyJsonVisitor {sender: sender.clone()}) {
[...]

PR #526, opened 2019-03-19
Add support for streaming a large JSON array

17dec:
My recent excursion into EOF error reporting made me realize that my hack (that happens to rely on `is_eof()`) to incrementally parse a large JSON array isn't the most robust approach, so here's my initial attempt in implementing a proper streaming array parser.

This implements issue #404 by adding an ArrayDeserializer API that mimics the StreamDeserializer. I did investigate extending the StreamDeserializer API itself, but a new struct seemed more natural.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -155,6 +155,15 @@ impl<'de, R: Read<'de>> Deserializer<R> {
         }
     }
 
+    /// Parse the JSON array as a stream of values.
+    pub fn into_array(self) -> ArrayDeserializer<'de, R> {
+        ArrayDeserializer {
+            de: self,
+            started: false,
+            lifetime: PhantomData,
+        }
+    }
+
     /// Parse arbitrarily deep JSON structures without any consideration for
     /// overflowing the stack.
     ///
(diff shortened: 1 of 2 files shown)


17dec:
Thanks for the feedback! I hadn't considered that use case. Unfortunately, that may complicate the API a little bit. Here's what I have in mind:

- Modify ArrayDeserializer to provide a return type generic next() implementation instead of an Iterator
- Rename `Deserializer::into_array_iter()` to `into_array()` - since it's not an Iterator anymore.
- Add an `ArrayDeserializer::iter()` (or `into_iter()` if I can't get borrowing to work) to still provide a convenient Iterator API for uniform element types. This step is optional, and I'm not sure the convenience is even that important.

How does this sound?

17dec:
Just pushed a new patch which implements points (1) and (2) of my proposal. I've not implemented the separate Iterator object (3) - this API is not likely to be used very often, so keeping the overall API small and simple may be preferable over the bit of extra convenience.

alexwennerberg:
Really interested in this PR! Is there anything preventing it from being merged at the moment?

dtolnay:
I wanted to brainstorm some alternatives but haven't been able to make time yet. Mainly I wonder how this API would be different if we wanted to support streaming nested arrays inside of arrays, arrays inside of objects, objects inside of objects, etc. Almost like a deserialization version of serde_json::ser::Formatter.

Maybe this could be experimented with in its own crate outside of serde_json, then considered for a PR?

17dec:
I've considered that use case, and think my proposed API can be extended quite naturally to support it. Fortunately, JSON only has two nested types, so it ought to be pretty simple (albeit a bit crude):

- `Deserializer::into_array()` -> `ArrayDeserializer`
- `Deserializer::into_map()` -> `ObjectDeserializer`
- `ArrayDeserializer::next_array()` -> `ArrayDeserializer`
- `ArrayDeserializer::next_map()` -> `ObjectDeserializer`
- `ObjectDeserializer::next_array()` -> `ArrayDeserializer`
- `ObjectDeserializer::next_map()` -> `ObjectDeserializer`

To make this work, `ObjectDeserializer` and `ArrayDeserializer` would need to have a lifetime so that one can't call `next()` on the originating struct while dealing with a nested array/object.

But I won't have time to work this out in the foreseeable future.

OscarRL:
> I've considered that use case, and think my proposed API can be extended quite naturally to support it. Fortunately, JSON only has two nested types, so it ought to be pretty simple (albeit a bit crude):
> 
> * `Deserializer::into_array()` -> `ArrayDeserializer`
> * `Deserializer::into_map()` -> `ObjectDeserializer`
> * `ArrayDeserializer::next_array()` -> `ArrayDeserializer`
> * `ArrayDeserializer::next_map()` -> `ObjectDeserializer`
> * `ObjectDeserializer::next_array()` -> `ArrayDeserializer`
> * `ObjectDeserializer::next_map()` -> `ObjectDeserializer`
> 
> To make this work, `ObjectDeserializer` and `ArrayDeserializer` would need to have a lifetime so that one can't call `next()` on the originating struct while dealing with a nested array/object.
> 
> But I won't have time to work this out in the foreseeable future.

This one should be the best approach. Actually as our ingestion pipeline we are handling JSON's like the one that you are talking here. But also we have cases where there is a simple BIG Json.

Big Json-Array proposed here:

```
[
  { .. element.. },
  ... GB of elements...
]
```

Alternative big JSON-Object used to transfer multiple data with the same file:

```
{
   "data1": [.... GB of elements...],
   "data2": [.... more GB of elements...]
}
```


JSON-Lines that can be parsed easy reading lines with a byte buffer and send to the lib.

```
[element]
[element]
... GB of elements...
```


Actually the lib always expects that the data fits in memory or it forces to the programmer build a custom deserializer and pop the data in the nested elements using a visitor pattern.

nwoltman:
Is it possible to implement the code in this PR outside of `serde_json`? I need the `ArrayDeserializer` for a project I'm working on but it uses methods that are marked as private, so it seems I can't use the code in my own crate.

(Sorry if there's an easy solution to this. It's only my second day using Rust.)

dtolnay:
Closing because it doesn't look like there is still design work ongoing and I don't plan to accept the current design. I would recommend pursuing a design that takes into account https://github.com/serde-rs/json/pull/526#pullrequestreview-216428907 and https://github.com/serde-rs/json/pull/526#issuecomment-505547968, whether in an external crate that builds on serde_json or in a fork. The examples in https://github.com/serde-rs/json/pull/526#issuecomment-519083982 would be a good test case to work toward getting working.

PR #527, opened 2019-03-22
Add hashbrown as option for serde_json::Value::Map

Licenser:
Using hashbrown for serde_json::Value gives a up to 10% improvement
in serialization/deserialization performance. On the flip side it
looses the option of a double ended itterator. This adds it as a
feature flag allowing people to choose it if it should suite their
needs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,12 @@ serde = "1.0.60"
 indexmap = { version = "1.0", optional = true }
 itoa = "0.4.3"
 ryu = "0.2"
-hashbrown = "*"
+
+# Use a different representation for the map type of serde_json::Value.
+# This improves performance but will not imoplement a DoubleEndedItterator.
+# This feature is incompatible with preserve_order.
+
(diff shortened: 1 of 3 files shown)


Licenser:
The benchmarks ignore that applications do not simply parse and stringify json, they do something with the data.

Licenser:
I think the difference you saw is based on the system allocator, runniung it w/ jemalloc returns the results:
```
target/release/json-benchmark-hashbrown
                                DOM                  STRUCT
data/canada.json         220 MB/s   350 MB/s
data/citm_catalog.json   340 MB/s   370 MB/s
data/twitter.json        260 MB/s   530 MB/s
```

```
heinz@Schrodinger ~/Projects/json-benchmark (master) $ target/release/json-benchmark-btree
                                DOM                  STRUCT
data/canada.json         220 MB/s   360 MB/s
data/citm_catalog.json   310 MB/s   310 MB/s
data/twitter.json        230 MB/s   480 MB/s
```

PR #534, opened 2019-04-11
fix outdated reference to serde macro location

Velfi:
The `Serialize` and `Deserialize` macros are no longer part of `serde` and have been moved to `serde_derive`

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ largely automatically.
 </a>
 
 ```rust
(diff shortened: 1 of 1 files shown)


Velfi:
You are totally right. I hadn't realized that they moved derives back into `serde` but under a feature flag. Thank you for taking the time to review this. I'm just going to close this.

PR #541, opened 2019-05-24
perfect_float feature for deserialization accuracy

### LINKED ISSUE #536: Perfect accuracy float parsing
Float parsing is currently implemented by calculating the significand as u64, casting to f64, and then multiplying or dividing by a nonnegative power of 10. For example the input `10.9876543210987654` would be parsed into the value `109876543210987654_u64 as f64 / 10e15`.

This algorithm is sometimes correct, or else usually close to correct in practical usage. It matches how JSON parsers are implemented in other languages.

However, it can happen that the result from this algorithm is not the mathematically nearest 64-bit floating point value to the exact value of the input. A "correct" algorithm would always produce the mathematically nearest answer. This requires high precision big-integer arithmetic in the general case so there would be a large performance cost; if implemented, we would likely want this behind a cfg that is off by default, with the current approximate behavior as default. This way programs can opt in to the more expensive algorithm as required.

```rust
fn main() {
    let input = "10.9876543210987654";
    let n: f64 = serde_json::from_str(input).unwrap();

    // produces 10.9876543210987644982878919108770787715911865234375
    // which is low by 9.017e-16
    let current_algorithm = 109876543210987654_u64 as f64 / 10e15;
    println!("{}", precise::to_string(current_algorithm));
    assert_eq!(n, current_algorithm);

    // produces 10.98765432109876627464473131112754344940185546875
    // which is high by 8.746e-16 (closer)
    let correct_answer = 10.9876543210987654_f64;
    println!("{}", precise::to_string(correct_answer));
    assert_ne!(n, correct_answer);
}
```

ijl:
This parses floats and integers using lexical-core from
https://github.com/AlexHuszagh/rust-lexical.

This fixes #536.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ appveyor = { repository = "serde-rs/json" }
 serde = { version = "1.0.60", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = { version = "0.4.3", default-features = false }
+lexical-core = { version = "0.7", optional = true, features = ["format"] }
 ryu = "1.0"
 
 [dev-dependencies]
@@ -62,6 +63,13 @@ preserve_order = ["indexmap"]
 # allows JSON numbers of arbitrary size/precision to be read into a Number and
 # written back to a JSON string without loss of precision.
 arbitrary_precision = []
+
+# Deserialize floats with perfect accuracy and consistent rounding. This
+# changes the implementation for parsing both floats and integers. This
+# implementation is slower than the default and contains a lookup table that
+# increases artifact size. It does not affect serialization. This
+# feature has no effect if arbitrary_precision is enabled.
+perfect_float = ["lexical-core"]
 
 # Provide a RawValue type that can hold unprocessed JSON during deserialization.
 raw_value = []
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -732,19 +732,9 @@ fn test_parse_number_errors() {
         ("0x80", "trailing characters at line 1 column 2"),
         ("\\0", "expected value at line 1 column 1"),
         ("1.", "EOF while parsing a value at line 1 column 2"),
-        ("1.a", "invalid number at line 1 column 3"),
-        ("1.e1", "invalid number at line 1 column 3"),
         ("1e", "EOF while parsing a value at line 1 column 2"),
         ("1e+", "EOF while parsing a value at line 1 column 3"),
         ("1a", "trailing characters at line 1 column 2"),
-        (
-            "100e777777777777777777777777777",
-            "number out of range at line 1 column 14",
-        ),
-        (
-            "-100e777777777777777777777777777",
-            "number out of range at line 1 column 15",
-        ),
         (
             "1000000000000000000000000000000000000000000000000000000000000\
              000000000000000000000000000000000000000000000000000000000000\
@@ -771,6 +761,38 @@ fn test_parse_number_errors() {
              000000000000000000000000000000000000000000000000000000000000\
              e9", // 1e309
             "number out of range at line 1 column 303",
+        ),
+    ]);
+
+    // the error messages here vary by implementation
+    test_parse_err::<f64>(&[
+        (
+            "1.a",
+            #[cfg(not(feature = "perfect_float"))]
+            "invalid number at line 1 column 3",
+            #[cfg(feature = "perfect_float")]
+            "EOF while parsing a value at line 1 column 2",
+        ),
+        (
+            "1.e1",
+            #[cfg(not(feature = "perfect_float"))]
+            "invalid number at line 1 column 3",
+            #[cfg(feature = "perfect_float")]
+            "EOF while parsing a value at line 1 column 4",
+        ),
+        (
+            "100e777777777777777777777777777",
+            #[cfg(not(feature = "perfect_float"))]
+            "number out of range at line 1 column 14",
+            #[cfg(feature = "perfect_float")]
+            "number out of range at line 1 column 31",
+        ),
+        (
+            "-100e777777777777777777777777777",
+            #[cfg(not(feature = "perfect_float"))]
+            "number out of range at line 1 column 15",
+            #[cfg(feature = "perfect_float")]
+            "number out of range at line 1 column 32",
         ),
     ]);
 }
@@ -840,14 +862,10 @@ fn test_parse_f64() {
         ("0.00e+00", 0.0),
         ("0.00e-00", 0.0),
         ("3.5E-2147483647", 0.0),
-        (
-            &format!("{}", (i64::MIN as f64) - 1.0),
-            (i64::MIN as f64) - 1.0,
-        ),
-        (
-            &format!("{}", (u64::MAX as f64) + 1.0),
-            (u64::MAX as f64) + 1.0,
-        ),
+        // (i64::MIN as f64) - 1.0
+        ("-9223372036854776000.0", -9223372036854776000.0),
+        // (u64::MAX as f64) + 1.0
+        ("18446744073709552000.0", 18446744073709552000.0),
         (&format!("{}", f64::EPSILON), f64::EPSILON),
         (
             "0.0000000000000000000000000000000000000000000000000123e50",
@@ -899,6 +917,28 @@ fn test_parse_f64() {
              000000000000000000e-10",
             1e308,
         ),
+    ]);
+
+    #[cfg(not(feature = "perfect_float"))]
+    #[cfg(not(feature = "arbitrary_precision"))]
+    test_parse_ok(vec![
+        (
+            // "-9223372036854776000", note it formats with no trailing ".0"
+            &format!("{}", (i64::MIN as f64) - 1.0),
+            (i64::MIN as f64) - 1.0,
+        ),
+        (
+            // "18446744073709552000", note it formats with no trailing ".0"
+            &format!("{}", (u64::MAX as f64) + 1.0),
+            (u64::MAX as f64) + 1.0,
+        ),
+    ]);
+
+    #[cfg(not(feature = "arbitrary_precision"))]
+    #[cfg(feature = "perfect_float")]
+    test_parse_ok(vec![
+        ("31.245270191439438", 31.245270191439438),
+        ("121.48791951161945", 121.48791951161945),
     ]);
 }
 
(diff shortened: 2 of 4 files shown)


ijl:
Benchmarks using https://github.com/serde-rs/json-benchmark:

master:
```
                                DOM                  STRUCT
data/canada.json         290 MB/s   540 MB/s   700 MB/s   430 MB/s
data/citm_catalog.json   450 MB/s   720 MB/s  1040 MB/s  1030 MB/s
data/twitter.json        320 MB/s  1000 MB/s   680 MB/s  1050 MB/s
```

perfect-float:
```
                                DOM                  STRUCT
data/canada.json         150 MB/s   490 MB/s   220 MB/s   370 MB/s
data/citm_catalog.json   420 MB/s   490 MB/s   860 MB/s   650 MB/s
data/twitter.json        310 MB/s   720 MB/s   650 MB/s   780 MB/s
```

tests/test.rs
```diff
@@ -783,6 +771,45 @@ fn test_parse_number_errors() {
...
+        (
+            "1e+",
+            #[cfg(not(feature = "perfect_float"))]
+            "EOF while parsing a value at line 1 column 3",
+            #[cfg(feature = "perfect_float")]
+            "invalid number at line 1 column 3",
+        ),
+    ]);
+
+    // these parse ok with perfect_float
```
dtolnay: This is problematic because they are not valid JSON.

tests/test.rs
```diff
@@ -909,6 +928,32 @@ fn test_parse_f64() {
              000000000000000000e-10",
             1e308,
         ),
+        // (i64::MIN as f64) - 1.0
+        ("-9223372036854776000.0", -9223372036854776000.0),
+        // (u64::MAX as f64) + 1.0
+        ("18446744073709552000.0", 18446744073709552000.0),
+    ]);
+
+    #[cfg(not(feature = "perfect_float"))]
```
dtolnay: How come these don't pass with perfect_float?

dtolnay:
@ijl is it going to be possible to fix the cases where perfect_float successfully parses something that is not valid JSON?

ijl:
@dtolnay I would need to look when I have more time. I can close this as stale until then if you prefer.

Alexhuszagh:
Let me know if I can be any help on integrating these changes to serde-json.

tests/test.rs
Alexhuszagh: @dtolnay The issue is because that value actually isn't the correct representation of a perfect float. That is a close-to-halfway representation, and the correct bit-pattern is actually:

```
0100001111110000000000000000000000000000000000000000000000000000
```

The annotated bit-pattern is:

| Sign  |  Exponent | Significant Digits |
|:-------:|:-------------:|:----------------------:|
|   0     | 10000111111 | 0000000000000000000000000000000000000000000000000000 |

As you can see, the significant digits has a 0 for the significant bit, so for anything below or equal to the halfway point (a bit set just past the significant digits), the default IEEE754 rounding scheme of round-nearest-tie-even will round down. The correct value is `18446744073709551616.0`.

Alexhuszagh: For simple code to practice getting halfway representations, use the following code:

```python
import numpy as np

class FloatMixin:
    '''Mixing for floating-point methods.'''

    def __init__(self, value):
        self.value = self.float_type(value)

    def to_bits(self):
        '''Extract bitwise representation of float.'''
        return np.frombuffer(self.value.tobytes(), dtype=self.int_type)[0]

    @classmethod
    def from_bits(cls, value):
        '''Extract bitwise representation of float.'''
        return cls(np.frombuffer(value.tobytes(), dtype=cls.float_type)[0])

    def to_hex(self):
        '''Convert double to hex.'''
        return '{0:0{1}x}'.format(self.to_bits(), self.value.itemsize * 2)

    def is_denormal(self):
        '''Returns true if the float is a denormal.'''
        return self.to_bits() & self.EXPONENT_MASK == 0

    def is_special(self):
        '''Returns true if the float is NaN or Infinite.'''
        return self.to_bits() & self.EXPONENT_MASK == self.EXPONENT_MASK

    def is_nan(self):
        '''Returns true if the float is NaN.'''
        return self.is_special() and self.to_bits() & self.MANTISSA_MASK != 0

    def is_inf(self):
        '''Returns true if the float is Infinite.'''
        return self.is_special() and self.to_bits() & self.MANTISSA_MASK == 0

    def exponent(self):
        '''Get exponent component from the float.'''

        if self.is_denormal():
            return self.DENORMAL_EXPONENT
        bits = self.to_bits()
        exp_bits = bits & self.EXPONENT_MASK
        biased_e = np.int32(exp_bits >> int_type(self.MANTISSA_SIZE))
        return biased_e - self.EXPONENT_BIAS

    def mantissa(self):
        '''Get mantissa component from the float.'''

        bits = self.to_bits()
        s = bits & self.MANTISSA_MASK
        if not self.is_denormal():
            return s + self.HIDDEN_BIT_MASK
        return s


class Float32(FloatMixin):
    '''Wrapper around a 32-bit floating point value.'''

    SIGN_MASK           = np.uint32(0x80000000)
    EXPONENT_MASK       = np.uint32(0x7F800000)
    HIDDEN_BIT_MASK     = np.uint32(0x00800000)
    MANTISSA_MASK       = np.uint32(0x007FFFFF)
    MANTISSA_SIZE       = np.int32(23)
    EXPONENT_BIAS       = np.int32(127 + MANTISSA_SIZE)
    DENORMAL_EXPONENT   = np.int32(1 - EXPONENT_BIAS)
    float_type = np.float32
    int_type = np.uint32


class Float64(FloatMixin):
    '''Wrapper around a 64-bit floating point value.'''

    SIGN_MASK           = np.uint64(0x8000000000000000)
    EXPONENT_MASK       = np.uint64(0x7FF0000000000000)
    HIDDEN_BIT_MASK     = np.uint64(0x0010000000000000)
    MANTISSA_MASK       = np.uint64(0x000FFFFFFFFFFFFF)
    MANTISSA_SIZE       = np.int32(52)
    EXPONENT_BIAS       = np.int32(1023 + MANTISSA_SIZE)
    DENORMAL_EXPONENT   = np.int32(1 - EXPONENT_BIAS)
    float_type = np.float64
    int_type = np.uint64
```

Now, to calculate `b+1`, the next float float from a given float `b`, and `b+h`, the halfway point, do:

```python
b = Float64(18446744073709551616.0)
b1 = Float64.from_bits(b.to_bits() + Float64.int_type(1))
bh = (int(b.value) + int(b1.value)) // 2
```

In this case, for `b` of `18446744073709551616.0`, `b+1` is `18446744073709555712.0`, and `b+h` is `18446744073709553664.0`. `b+h` cannot be represented as an exact floating-point number, but we can confirm it is accurate with a few tests:

```python
>>> float("18446744073709553664.0")  # b+h
18446744073709551616.0
>>> float("18446744073709553665.0")  # above b+h
18446744073709555712.0
```

Alexhuszagh: @dtolnay The accurate float parser does not mean every decimal representation can be exactly represented as a floating-point number, however, it does mean that the closest representation will always be used, and that these conversions are stable.

I hope my explanation makes sense?

Cargo.toml
```diff
@@ -20,6 +20,7 @@ appveyor = { repository = "serde-rs/json" }
 serde = { version = "1.0.60", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = { version = "0.4.3", default-features = false }
+lexical-core = { version = "0.7", optional = true, default_features = false, features = ["std", "correct", "format", "ryu"]}
```
Alexhuszagh: You can just use `default_features = true`, features = ["format"]`, since the default features are `std`, `ryu`, and `correct` always,

Cargo.toml
```diff
@@ -63,6 +64,13 @@ preserve_order = ["indexmap"]
 # written back to a JSON string without loss of precision.
 arbitrary_precision = []
 
+# Deserialize floats with perfect accuracy and consistent rounding. This
+# changes the implementation for parsing both floats and integers. This
+# implementation is slower than the default and contains a lookup table that
+# increases artifact size. It does not affect serialization. This
+# feature has no effect if arbitrary_precision is enabled.
+perfect_float = ["lexical-core"]
```
Alexhuszagh: I would recommend using `perfect_float` by default. @dtolnay will ultimately decide, however. Correctness should always be favored by default, and performance at the cost of correctness should be opt-in, not opt-out.

ijl:
I've pushed what I had been using, fixed up for stable and incompatibility with rust 1.31. This could be reviewed again.

As I added in the amended commit message, avoiding copying to scratch is an obvious thing to do. It's a larger change, though.

tests/test.rs
ijl: This has since been fixed.

Cargo.toml
ijl: Ok, thanks.

Alexhuszagh:
@ijl I might recommend using v0.6 rather than v0.7, since v0.7 makes a breaking change and removes support for Rustc < `1.37`. v0.6 supports Rustc >= `1.24`, which is ages.

tests/test.rs
dtolnay: I understand that in general not every decimal representation can be exactly represented as a floating-point number, that is obvious to me. The part I don't think you answered is why these particular values in the test no longer round trip.

My expectation for f64::to_string is that it produces a decimal representation (hopefully with as few significant digits as possible, but this is not necessary) which is mathematically closer to the original input float than to any other float.

My expectation for f64::from_str is that it produces the float which is mathematically closest to the input digits.

This test was supposed to test that pair of expectations. Of the two expectations, is one or both untrue of the standard library? Is one or both untrue of the implementation in this PR?

For the standard library my understanding is that the following line succeeds for any input float `n`, and that's certainly true for the two specific floats in this test.

```rust
assert_eq!(n.to_string().parse::<f64>().unwrap(), n);
```

Alexhuszagh: @dtolnay Oh I see what you mean, my mistake. However, the actual results seem to work after adding the test back in anyway:

```rust
    #[cfg(not(feature = "arbitrary_precision"))]
    #[cfg(feature = "perfect_float")]
    test_parse_ok(vec![
        ("18446744073709552000.0", 18446744073709552000.0),
        ("31.245270191439438", 31.245270191439438),
        ("121.48791951161945", 121.48791951161945),
    ]);
```

ijl:
In #536 it looks like @Alexhuszagh will create a minimal float parser from lexical for serde_json, so I'm happy to close this.

PR #544, opened 2019-05-30
Add note about issue 160 to from_reader function

est31:
Today I stumbled across the fact that from_reader is slow. There seems to be a suggestion from back in 2016 to add a comment to the docs https://github.com/serde-rs/json/issues/160#issuecomment-258589916 but it seems it was never done. This PR finally adds that note as apparently people like me still stumble across the bug. See #160.


est31:
ping @dtolnay  @oli-obk @erickt wanna review?

PR #558, opened 2019-08-17
add no_std support

Freax13:
I really needed JSON serde support for a no_std project. I managed to add two new features "std" and "no_std" to support both std and no_std/alloc environments. I know that there is a serde_json_core crate, but this crate doesn't offer full json support such as escapes values.

Usually, one only adds one feature for "std" and has that be enabled by default, but in this specific case, I decided to add a no_std feature because I needed to depend on core_io for no_std support (AFAIK there is no way to conditionally disable a dependency with a feature hence the no_std feature).

Due to the way core_io works, this will only compile one some specific compiler versions (the latest one being nightly-2019-07-01)

I should also note that the tests can only be run in std mode. Some of the dev dependencies depend on serde with std enabled and that causes std to be enabled globally. This only affects the tests though, if this crate is used in another crate as a dependency the dev dependencies don't kick in.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,27 +1,25 @@
 //! When serializing or deserializing JSON goes wrong.
 
-#[cfg(not(feature = "no_std"))]
-use std::error;
-#[cfg(not(feature = "no_std"))]
-use std::fmt::{self, Debug, Display};
-#[cfg(feature = "no_std")]
-use core::fmt::{self, Debug, Display};
-#[cfg(not(feature = "no_std"))]
-use std::io;
-#[cfg(feature = "no_std")]
-use core_io as io;
-#[cfg(not(feature = "no_std"))]
-use std::result;
-#[cfg(feature = "no_std")]
-use core::result;
-#[cfg(not(feature = "no_std"))]
-use std::str::FromStr;
-#[cfg(feature = "no_std")]
-use core::str::FromStr;
-#[cfg(feature = "no_std")]
+#[cfg(not(feature = "std"))]
 use alloc::boxed::Box;
-#[cfg(feature = "no_std")]
+#[cfg(not(feature = "std"))]
 use alloc::string::{String, ToString};
+#[cfg(not(feature = "std"))]
+use core::fmt::{self, Debug, Display};
+#[cfg(not(feature = "std"))]
+use core::result;
+#[cfg(not(feature = "std"))]
+use core::str::FromStr;
+#[cfg(feature = "std")]
+use std::error;
+#[cfg(feature = "std")]
+use std::fmt::{self, Debug, Display};
+#[cfg(feature = "std")]
+use std::io;
+#[cfg(feature = "std")]
+use std::result;
+#[cfg(feature = "std")]
+use std::str::FromStr;
 
 use serde::de;
 use serde::ser;
@@ -67,6 +65,7 @@ impl Error {
     pub fn classify(&self) -> Category {
         match self.err.code {
             ErrorCode::Message(_) => Category::Data,
+            #[cfg(feature = "std")]
             ErrorCode::Io(_) => Category::Io,
             ErrorCode::EofWhileParsingList
             | ErrorCode::EofWhileParsingObject
@@ -148,6 +147,7 @@ pub enum Category {
 }
 
 #[cfg_attr(feature = "cargo-clippy", allow(fallible_impl_from))]
+#[cfg(feature = "std")]
 impl From<Error> for io::Error {
     /// Convert a `serde_json::Error` into an `io::Error`.
     ///
@@ -176,7 +176,7 @@ impl From<Error> for io::Error {
     ///     }
     /// }
     /// ```
-    #[cfg(not(feature = "no_std"))]
+    #[cfg(feature = "std")]
     fn from(j: Error) -> Self {
         if let ErrorCode::Io(err) = j.err.code {
             err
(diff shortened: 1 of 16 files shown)


Freax13:
almost everything because ser::Serializer, de::Deserializer, to_vec and from_str all depend on io::Write or io::Read

If that's a huge issue we could try rewriting/copying small parts of std::io for no_std mode instead of depending on core_io.

dtolnay:
But to_vec and from_str don't require io. It seems easily possible to implement those signatures without std::io, only alloc.

https://docs.serde.rs/serde_json/fn.to_vec.html
https://docs.serde.rs/serde_json/fn.from_str.html

Could you try that approach? We would have only a "std" feature and that feature would be purely additive, adding only those public APIs that can't exist without std::io.

dtolnay:
Even Deserializer can be provided without io, only the Deserializer::from_reader method would need to be std-only.

Freax13:
but to_vec and from_str depend on Serializer and Deserializer which require io. Deserializer also strongly depends on std::io. even methods that don't mention io like from_slice internally convert the slice to a reader and use that to create a deserializer

the only way i can see to get rid of core_io is to rewrite/copy parts from std::io

dtolnay:
What I mean is that it is possible to implement the exact public API of to_vec and from_str and nearly all of the public API of Deserializer using some alternative implementation when not using std. That's the direction that I would prefer for this to go, rather than using core_io or vendoring parts of core_io.

As far as I can tell none of the implementation of Deserializer requires std::io outside of the from_reader constructor which can be std-only.

Freax13:
I see. I was confusing read::Read with io::Read. I'll give it a try

Freax13:
okay, so that was way easier than I thought it would be. I created an abstraction over Write just like you already did with Read in serde_json::read and used that in all methods. I completely removed the no_std feature

Freax13:
hey, so I fixed all build errors, let me know if there's anything else you'd want me to change

src/error.rs
```diff
@@ -159,6 +176,7 @@ impl From<Error> for io::Error {
     ///     }
     /// }
     /// ```
+    #[cfg(feature = "std")]
```
dtolnay: This is redundant because the impl block already has the same cfg.

src/error.rs
```diff
@@ -170,6 +188,21 @@ impl From<Error> for io::Error {
             }
         }
     }
+
+    #[cfg(not(feature = "std"))]
+    fn from(j: Error) -> Self {
```
dtolnay: This function is never compiled. The surrounding impl block only exists if std is enabled. Please remove.

src/lib.rs
```diff
@@ -319,27 +319,39 @@
...
+pub use self::ser::{to_string, to_string_pretty, to_vec, to_vec_pretty, Serializer};
+#[cfg(feature = "std")]
+pub use self::ser::{to_writer, to_writer_pretty};
 #[doc(inline)]
 pub use self::value::{from_value, to_value, Map, Number, Value};
 
+#[cfg(not(any(feature = "std", feature = "alloc")))]
+compile_error!("std and alloc are both disabled");
+#[cfg(all(feature = "std", feature = "alloc"))]
+compile_error!("std and alloc are both enabled");
```
dtolnay: This needs to not be an error. Please find a way to support the case where the same dependency graph contains one crate that only needs serde_json with alloc and a different crate needs serde_json with std. That's part of what it means for features to be additive. I believe the serde crate currently does this correctly.

src/lib.rs
```diff
@@ -348,6 +360,15 @@ macro_rules! try {
         }
     };
 }
+#[cfg(not(feature = "std"))]
+macro_rules! try {
```
dtolnay: This duplication is not needed (here, and elsewhere in the PR in imports). It is fine for a normal std crate to import from core. If something is in core, you should avoid duplication by using it from core regardless of what features are set.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=ce95ea2938a54b82d24ff7859882b39a

src/ser.rs
```diff
@@ -1611,18 +1384,18 @@ impl CharEscape {
 pub trait Formatter {
     /// Writes a `null` value to the specified writer.
     #[inline]
-    fn write_null<W: ?Sized>(&mut self, writer: &mut W) -> io::Result<()>
+    fn write_null<W: ?Sized>(&mut self, writer: &mut W) -> Result<()>
```
dtolnay: This is a breaking change. Please avoid making this change.

The Formatter trait should not be included in the public API in no_std mode.

src/value/partial_eq.rs
```diff
@@ -59,6 +61,7 @@ impl PartialEq<Value> for String {
 macro_rules! partialeq_numeric {
     ($($eq:ident [$($ty:ty)*])*) => {
         $($(
+            #[cfg(feature = "std")]
```
dtolnay: Why does this impl need to require std?

src/ser.rs
```diff
@@ -19,7 +34,7 @@ pub struct Serializer<W, F = CompactFormatter> {
 
 impl<W> Serializer<W>
 where
-    W: io::Write,
+    W: write::Write,
```
dtolnay: This is a breaking change. Please don't change the public API of Serializer in std mode.

dtolnay:
Hi @Freax13 -- are you still interested in continuing to work on this PR?

PR #561, opened 2019-08-20
Return None instead of lossy f64 conversion

Wolvereness:
Most numbers representable by `u64` and `i64` aren't representable as `f64` starting from `+/- 9007199254740993`. As per documentation, these should return `None`.


Wolvereness:
> Represents the number as f64 if possible. Returns None otherwise.

The documentation is **plainly wrong**. It is not possible to represent most of the numbers outside of that range. Whatever *both sides* argument there is, I staunchly object to a mismatch between documentation and the code.

PR #585, opened 2019-11-19
Add `impl From<()> for Value`

Nilix007:
Hey there,

this tiny PR adds the conversion from `()` to `Value::Null`.

I feel this is justified as serde_json already perceives `()` as the JSON value `null`:
```
println!("{:?}", serde_json::from_str::<()>("null"));
// prints `Ok(())`
    
println!("{:?}", serde_json::to_string(&()));
// prints `Ok("null")`
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0eac935756f3537acdf45486dcabb4db

Cheers,
Felix

PS: Please have a look at the marvelous signature of `<Value as From<()>>::from` :laughing:

### DIFF
--- a/src/value/from.rs
+++ b/src/value/from.rs
@@ -211,3 +211,19 @@ impl<T: Into<Value>> ::std::iter::FromIterator<T> for Value {
         Value::Array(iter.into_iter().map(Into::into).collect())
     }
 }
+
+impl From<()> for Value {
+    /// Convert `()` to `Value`
+    ///
+    /// # Examples
+    ///
+    /// ```edition2018
+    /// use serde_json::Value;
+    ///
+    /// let u = ();
+    /// let x: Value = u.into();
+    /// ```
+    fn from((): ()) -> Self {
+        Value::Null
+    }
+}


Nilix007:
Mh. So I was working on a library which uses diesel's [`Jsonb`](https://docs.rs/diesel/1.4.3/diesel/pg/types/sql_types/struct.Jsonb.html) internally for some kind of dynamic data field.  As `Jsonb` maps directly to `Value`, I'm exporting functions like `fn do_something<T: Into<Value>>(additional_data: T) {}` so that I can avoid pulling in the complete serde dependency tree and skip the serialization step completely (yay, one error case less on my side due to type safety).

Anyway, long story short, I think that the conversion from this PR would be quite convenient for the end user of my lib. And as it kinda feels "natural" to me, I sent this PR.

dtolnay:
What do you mean by "pulling in the complete serde dependency tree"? If you depend on serde_json to use Value, that depends on serde anyway.

Do you prefer Into\<Value\> over:

```rust
fn do_something(additional_data: impl Serialize)
```

which would be more flexible?

Nilix007:
> What do you mean by "pulling in the complete serde dependency tree"? If you depend on serde_json to use Value, that depends on serde anyway.
> 
Err, you're right, of course.  Please ignore this part :grimacing: 

> Do you prefer Into<Value> over:
> 
> ```rust
> fn do_something(additional_data: impl Serialize)
> ```
> 
> which would be more flexible?

Yes, because serde's serialization (and `serde_json::to_value` respectively) can fail while From/Into cannot.

dtolnay:
Published in 1.0.42.

Issue #505, opened 2018-11-28
Arbitrary precision numbers don't work with `serde(tag = ..)` and `serde(flatten)`


idubrov:
The code below fails if `arbitrary_precision` feature is turned on (works fine if it is turned off):

```rust
#[macro_use]
extern crate serde_derive;

#[derive(Deserialize)]
struct Data {
    value: i32,
}

#[derive(Deserialize)]
#[serde(tag = "type")]
enum Wrapper {
    Data(Data),
}

fn main() {
    let json = r#"{"type":"Data","value":123}"#;
    // Okay
    let _data1: Data = serde_json::from_str(json).unwrap();
    // Fails!
    let _data2: Wrapper = serde_json::from_str(json).unwrap();
}
```

Error backtrace:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: map, expected i32", line: 0, column: 0)', libcore/result.rs:1009:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
```

Cargo.toml:

```toml
[package]
name = "serde-arbitrary-fail"
version = "0.1.0"
authors = ["Ivan Dubrov <dubrov.ivan@gmail.com>"]

[dependencies]
serde = "1.0.80"
serde_derive = "1.0.80"
serde_json = { version = "1.0.33", features = ["arbitrary_precision"]}
```

igorbernstein:
It also fails for `#[serde(untagged)]`
```
data did not match any variant of untagged enum Wrapper
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum Wrapper", line: 0, column: 0)', src/libcore/result.rs:997:5
stack backtrace:
```

idubrov:
Same for `#[flatten]`:

```rust
use serde_derive::{Serialize, Deserialize};

#[derive(Debug, Serialize, Deserialize)]
struct Outer {
    #[serde(flatten)]
    inner: Inner,
}

[...]

PR #586, opened 2019-11-21
Fix number deserialization with `arbitrary_precision`

bkchr:
When we have a struct that uses `#[serde(flatten)]` and the feature
`arbitrary_precision` is enabled, the deserialization. The problem was
that the arbitrary number was forwarded as a "map" to the visitor and in
a later step this map failed as a number was expected.

Fixes: https://github.com/serde-rs/json/issues/505

Requires: https://github.com/serde-rs/serde/pull/1679

### DIFF
--- a/tests/test.rs
+++ b/tests/test.rs
@@ -956,21 +956,6 @@ fn test_parse_number() {
         ("\t1.0", "invalid number at line 1 column 1"),
         ("1.0\t", "invalid number at line 1 column 4"),
     ]);
-
-    #[cfg(feature = "arbitrary_precision")]
-    test_parse_ok(vec![
-        ("1e999", Number::from_string_unchecked("1e999".to_owned())),
-        ("-1e999", Number::from_string_unchecked("-1e999".to_owned())),
-        ("1e-999", Number::from_string_unchecked("1e-999".to_owned())),
-        (
-            "2.3e999",
-            Number::from_string_unchecked("2.3e999".to_owned()),
-        ),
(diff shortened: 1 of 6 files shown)


tests/test.rs
```diff
@@ -956,21 +956,6 @@ fn test_parse_number() {
         ("\t1.0", "invalid number at line 1 column 1"),
         ("1.0\t", "invalid number at line 1 column 4"),
     ]);
-
-    #[cfg(feature = "arbitrary_precision")]
```
bkchr: I removed this test + the extra added function `from_string_unchecked`. In my opinion this test does not makes that much sense at all. Even if we would have parsed the `1e999` correctly into the `Number`, there is not way to get this value out of the object.

WiSaGaN:
The patch seems still failing the test, which would have been passed without "arbitrary_precision" feature.
```rust
use serde::{Deserialize};

#[derive(Deserialize)]
#[serde(tag = "tag")]
#[derive(Clone, Debug, PartialEq)]
pub enum E1 {
    A{v: f64},
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn serde_1() {
        let incoming_str = r#"{"tag":"A","v":0}"#;
        let expected_incoming = E1::A{ v: 0.0 };
        let incoming = serde_json::from_str(&incoming_str).expect("cannot deserialize incoming_str");
        assert_eq!(expected_incoming, incoming);
    }
}
```

PR #590, opened 2019-12-04
Fix Required Minimum `indexmap` Version

nlordell:
The `preserve-order` feature uses `indexmap::map::OccupiedEntry::swap_remove` method which was only added in `1.2`, however, the `Cargo.toml` incorrectly requires `^1.0`.


dtolnay:
Published in 1.0.43.

PR #591, opened 2019-12-04
Implement IntoDeserializer for Value

erickt:
This allows users to deserialize from types like `HashMap<String, Value>`.


dtolnay:
Published in serde_json 1.0.44.

PR #598, opened 2020-01-03
Added more raw functions

Koxiaet:
This PR just adds some more useful functions for using RawValues: unsafe functions for converting to RawValues directly from a str, and `to_raw` which converts a Serializable object into a RawValue.

Things to think about:
* I put `to_raw` in both the crate root and the `value` module like `from_value` and `to_value`, but since `RawValue` is in the `value` module, perhaps it should be in the `value` module.
* I didn't implement `from_raw`, as all it is is `get` + `from_str`, and I had some trouble with its lifetimes in the tests.

I also had to make some changes to other parts of the code to get clippy to compile it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -344,6 +344,8 @@ pub use self::ser::{
 };
 #[doc(inline)]
 pub use self::value::{from_value, to_value, Map, Number, Value};
+#[cfg(feature = "raw_value")]
+pub use self::value::{from_raw, to_raw};
 
 // We only use our own error type; no need for From conversions provided by the
 // standard library's try! macro. This reduces lines of LLVM IR by 4%.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -345,7 +345,7 @@ pub use self::ser::{
 #[doc(inline)]
 pub use self::value::{from_value, to_value, Map, Number, Value};
 #[cfg(feature = "raw_value")]
-pub use self::value::{from_raw, to_raw};
+pub use self::value::to_raw;
 
 // We only use our own error type; no need for From conversions provided by the
 // standard library's try! macro. This reduces lines of LLVM IR by 4%.
--- a/src/number.rs
+++ b/src/number.rs
@@ -18,7 +18,7 @@ use error::ErrorCode;
 #[cfg(feature = "arbitrary_precision")]
 /// Not public API. Should be pub(crate).
 #[doc(hidden)]
-pub const TOKEN: &'static str = "$serde_json::private::Number";
+pub const TOKEN: &str = "$serde_json::private::Number";
 
 /// Represents a JSON number, whether integer or floating point.
 #[derive(Clone, PartialEq)]
@@ -746,6 +746,8 @@ impl Number {
     // Not public API. Should be pub(crate).
     #[doc(hidden)]
     #[cold]
+    // This function does not compile with clippy, despite the below allow.
+    #[allow(clippy::unused_self)]
     pub fn unexpected(&self) -> Unexpected {
         Unexpected::Other("number")
     }
--- a/src/number.rs
+++ b/src/number.rs
@@ -746,8 +746,6 @@ impl Number {
     // Not public API. Should be pub(crate).
     #[doc(hidden)]
     #[cold]
-    // This function does not compile with clippy, despite the below allow.
-    #[allow(clippy::unused_self)]
     pub fn unexpected(&self) -> Unexpected {
         Unexpected::Other("number")
     }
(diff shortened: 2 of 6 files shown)


src/raw.rs
```diff
@@ -117,7 +118,7 @@ pub struct RawValue {
 
 impl RawValue {
     fn from_borrowed(json: &str) -> &Self {
-        unsafe { mem::transmute::<&str, &RawValue>(json) }
+        unsafe { &*(json as *const str as *const RawValue) }
```
Koxiaet: Had to change this to get clippy to compile.

dtolnay: I would prefer not to make this change. I think this lint is bad and I have suppressed it in https://github.com/serde-rs/json/commit/88ee1a69dab73a45cd0eca5db505164fcd48cbf1. The original code is clearer and safer.

src/raw.rs
```diff
@@ -160,6 +161,14 @@ impl Display for RawValue {
     }
 }
 
+impl PartialEq<RawValue> for RawValue {
+    fn eq(&self, other: &RawValue) -> bool {
+        self.json.eq(&other.json)
+    }
+}
+
+impl Eq for RawValue {}
```
dtolnay: I don't think we should add these impls. What is your use case for them? It is potentially surprising that the same JSON object with only whitespace differences would compare unequal. Users who want this behavior should use `a.get() == b.get()` to compare the raw values as &amp;str.

src/number.rs
```diff
@@ -18,7 +18,7 @@ use error::ErrorCode;
 #[cfg(feature = "arbitrary_precision")]
 /// Not public API. Should be pub(crate).
 #[doc(hidden)]
-pub const TOKEN: &'static str = "$serde_json::private::Number";
+pub const TOKEN: &str = "$serde_json::private::Number";
```
dtolnay: In the future it would be better to make this kind of unrelated change in a separate commit in a separate PR. I've moved it to 57bc9ae77b31969ee6888899ca39f3e86ff0890e.

src/lib.rs
```diff
@@ -344,6 +344,8 @@ pub use self::ser::{
 };
 #[doc(inline)]
 pub use self::value::{from_value, to_value, Map, Number, Value};
+#[cfg(feature = "raw_value")]
+pub use self::value::to_raw;
```
dtolnay: Please remove this reexport. This is fine staying in the `value` module with RawValue.

src/raw.rs
```diff
@@ -454,3 +508,14 @@ impl<'de> MapAccess<'de> for BorrowedRawDeserializer<'de> {
         seed.deserialize(BorrowedStrDeserializer::new(self.raw_value.take().unwrap()))
     }
 }
+
+/// Convert a `T` into `serde_json::value::RawValue` which is a single valid
+/// JSON value.
+pub fn to_raw<T>(value: T) -> Result<Box<RawValue>, Error>
```
dtolnay: I think `to_raw_value` would be a clearer name. We had decided not to call the type just `Raw`.

Could you explain a real world use case where this function would be needed?

src/raw.rs
```diff
@@ -179,6 +188,35 @@ impl RawValue {
...
+    ///
+    /// # Safety
+    ///
+    /// This function assumes that:
+    ///
+    /// - the input is valid JSON, and
+    /// - the input has no leading or trailing whitespace.
+    pub unsafe fn from_str_unchecked(json: &str) -> &Self {
+        Self::from_borrowed(json)
+    }
```
dtolnay: I would like to avoid adding these functions. Do you have a real use case where these are essential?

src/raw.rs
```diff
@@ -215,9 +253,25 @@ impl RawValue {
...
+    /// Converts a `Box<RawValue>` into its JSON text as a boxed str without
+    /// copying or allocating.
+    pub fn into_boxed_str(self: Box<Self>) -> Box<str> {
+        unsafe { mem::transmute::<Box<Self>, Box<str>>(self) }
+    }
+    /// Converts a `Box<RawValue>` into its JSON bytes without copying or
+    /// allocating.
+    pub fn into_boxed_bytes(self: Box<Self>) -> Box<[u8]> {
+        self.into_boxed_str().into_boxed_bytes()
+    }
```
dtolnay: These seem redundant and niche. The caller can get these representations through APIs on String. Please remove.

Koxiaet:
This PR was created for a use case where I needed it, but I later realized that there was a much *much* better way to solve this problem. Thanks for taking the time to review it, but I will close this.

PR #605, opened 2020-01-20
Update MSRV to 1.30 in README.md

Xanewok:
As it seems to be the case as of latest commit (https://github.com/serde-rs/json/commit/4b8380261b63e0758877ab9cfe7ed1652cca4f88).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
-# Serde JSON &emsp; [![Build Status]][travis] [![Latest Version]][crates.io] [![Rustc Version 1.15+]][rustc]
+# Serde JSON &emsp; [![Build Status]][travis] [![Latest Version]][crates.io] [![Rustc Version 1.30+]][rustc]
 
 [Build Status]: https://api.travis-ci.org/serde-rs/json.svg?branch=master
 [travis]: https://travis-ci.org/serde-rs/json
 [Latest Version]: https://img.shields.io/crates/v/serde_json.svg
 [crates.io]: https://crates.io/crates/serde\_json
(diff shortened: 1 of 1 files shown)


Xanewok:
Also, I'm not sure if I'm doing things right but it seems that for development we pull serde_derive, which needs syn v1, which in turn only works on 1.31...
```
$ cargo +1.30.0 check
   Compiling proc-macro2 v1.0.8
   Compiling unicode-xid v0.2.0                                                                                                                                                                                                                                                                                               
   Compiling syn v1.0.14                                                                                                                                                                                                                                                                                                      
   Compiling ryu v1.0.2                                                                                                                                                                                                                                                                                                       
   Compiling serde v1.0.104                                                                                                                                                                                                                                                                                                   
    Checking itoa v0.4.4                                                                                                                                                                                                                                                                                                      
error: Edition 2018 is unstable and only available for nightly builds of rustc.                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
error: Could not compile `proc-macro2`.                                                                                                                                                                                                                                                                                       
warning: build failed, waiting for other jobs to finish...
error: Edition 2018 is unstable and only available for nightly builds of rustc.                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
error: Could not compile `syn`.                                                                                                                                                                                                                                                                                               
warning: build failed, waiting for other jobs to finish...
error: build failed                
```
Since we now support 1.30+, maybe we could bump up MSRV by one as well to bring dev and runtime version requirements to parity?

PR #606, opened 2020-01-20
Add a no_std/alloc feature

Xanewok:
This builds up on #588 (thanks @Freax13!).

With this, we support `no_std` with `alloc` crate, starting from Rust 1.36 (this was done so that we can use it in WebAssembly environment but any such environment will work).

To minimize the noise, I tried to use the facade approach already used in Serde to provide some consistency between the two and improved on the `io` facade to work around it missing in `core`.

Thanks in advance for taking your time to review this and let me know if I can improve anything to help this land!

Closes #588. Closes #516.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ appveyor = { repository = "serde-rs/json" }
 [dependencies]
 serde = { version = "1.0.60", default-features = false }
 indexmap = { version = "1.2", optional = true }
-itoa = "0.4.3"
+itoa = { version = "0.4.3", default-features = false }
 ryu = "1.0"
 
 [dev-dependencies]
@@ -44,7 +44,7 @@ features = ["raw_value"]
 [features]
 default = ["std"]
 
-std = ["serde/std"]
+std = ["serde/std", "itoa/std"]
 
 alloc = ["serde/alloc"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ travis-ci = { repository = "serde-rs/json" }
 appveyor = { repository = "serde-rs/json" }
 
 [dependencies]
-serde = "1.0.60"
+serde = { version = "1.0.60", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = "0.4.3"
 ryu = "1.0"
@@ -42,7 +42,11 @@ features = ["raw_value"]
 ### FEATURES #################################################################
 
 [features]
-default = []
+default = ["std"]
+
+std = ["serde/std"]
+
+alloc = ["serde/alloc"]
 
 # Use a different representation for the map type of serde_json::Value.
 # This allows data to be read into a Value and written back to a JSON string
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,10 @@ default = ["std"]
 
 std = ["serde/std", "itoa/std"]
 
+# Provide integration for heap-allocated collections without depending on the
+# rest of the Rust standard library.
+# NOTE: Disabling both `std` *and* `alloc` features is not supported yet.
+# Available on Rust 1.36+.
 alloc = ["serde/alloc"]
 
 # Use a different representation for the map type of serde_json::Value.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,7 +342,7 @@ extern crate ryu;
 
 ////////////////////////////////////////////////////////////////////////////////
 
-#[cfg(feature = "alloc")]
+#[cfg(not(feature = "std"))]
 extern crate alloc;
 
 /// A facade around all the types we need from the `std`, `core`, and `alloc`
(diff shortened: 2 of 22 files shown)


Cargo.toml
```diff
@@ -42,7 +42,15 @@ features = ["raw_value"]
...
 
 [features]
-default = []
+default = ["std"]
+
+std = ["serde/std", "itoa/std"]
+
+# Provide integration for heap-allocated collections without depending on the
+# rest of the Rust standard library.
+# NOTE: Disabling both `std` *and* `alloc` features is not supported yet.
```
dtolnay: Please find a way to produce a reasonable error message for this case. The current error if you do `serde_json = { default-features = false }` is \>1500 lines which is insane.

https://github.com/dtolnay/syn/blob/1.0.14/tests/features/mod.rs + https://github.com/dtolnay/syn/blob/1.0.14/tests/features/error.rs shows one good way.

src/lib.rs
```diff
@@ -334,6 +335,67 @@ extern crate indexmap;
...
+    pub use self::core::cell::{Cell, RefCell};
+    pub use self::core::clone::{self, Clone};
+    pub use self::core::convert::{self, From, Into};
+    pub use self::core::default::{self, Default};
+    pub use self::core::fmt::{self, Debug, Display};
+    pub use self::core::hash::{self, Hash};
+    pub use self::core::marker::{self, PhantomData};
+    pub use self::core::result::{self, Result};
+
+    #[cfg(all(feature = "alloc", not(feature = "std")))]
```
dtolnay: This can be simplified to `not(feature = "std")` because we know alloc must be enabled in that case. Same thing below.

In general `feature = "alloc"` should not need to appear anywhere.

src/lib.rs
```diff
@@ -365,6 +427,7 @@ pub mod map;
 pub mod ser;
```
dtolnay: From `cargo doc --no-default-features --features alloc`:

> Screenshot from 2020-01-20 23-26-49

<br>

This seems bad because the following would be accepted in alloc mode and break if a different part of the dependency graph enables std mode.

```rust
struct S;
impl serde_json::ser::Formatter for S {
    fn write_null<W: ?Sized>(&mut self, _: &mut W) -> Result<(), &'static str> {
        Err("")
    }
}
```

Either Formatter needs to be removed from the public API in alloc mode or the error type needs to be changed to an opaque type with no more than a subset of io::Error's API.

src/lib.rs
Xanewok: I did that to introduce distinction between `std`, `alloc` and no feature at all, but maybe it'd be good only introduce two: default `std` and no-`std` (which implies `alloc` and is supported from 1.36 onwards)...

I'll adapt the code accordingly, thanks!

Cargo.toml
Xanewok: Oh, that's neat! I wanted to do so but couldn't think of a reliable way to only show one compilation error.

As written [here](https://github.com/serde-rs/json/pull/606/files#r369071931), maybe it'd be good to just support `std` and no-`std` (implying using `alloc`)? This would mean we wouldn't have to care about detecting no `std` and `alloc` anymore.

src/lib.rs
Xanewok: Oops, I didn't notice that, thanks!

Since `std::io::Result` is part of the default, public API I don't think we can be backwards compatible (methods are one thing but sth may depend on it being explicitly this type from `std` IIUC).

Because of this, I think the only thing we can do now, until `std::io` somehow lands in `core`, is to hide this from public API in no-`std` case.

Xanewok: Reimplemented a small subset of `std::io::Error` and friends so that accidentally opting into `std` should not break stuff

Cargo.toml
Xanewok: Done in https://github.com/serde-rs/json/pull/606/commits/40b984cfcca926f2c0cc176b3bcf91b06fa6460f.

Is there a way to run trybuild with custom feature flags, like compiletest? It'd be quite handy to have a UI test for `--no-default-features` so that we can catch whenever the error message regresses.

Xanewok:
I addressed the feedback and thus also managed to unify the `io` facade as well (which now lives in `io/mod.rs`, while the small no_std-specific mplementation lives in `io/core.rs`). With this, the only place which specifies `std`-related conditional compilation is:
1. lib.rs - for the prelude/facade used across the library
2. io module - to reexport either `std::io` or the `no_std` reimplemented bits
3. features_check - to provide user-friendly compilation error for missing both std/alloc features

Overall, I'm pretty happy with how the `io` improvements turned out to be and with the overall patch. Thanks for reviewing this and let me know if I should change anything else!

Xanewok:
Awesome, thank you!

> for your use case do you require to_writer and/or Serializer, or only to_string?

I think `to_string` suffices for now, yeah.

> (...) and I also don't want to be in the business of copying many more Write impls from std into serde_json.

That's completely understandable! It's worth noting that for the second iteration of `io` I designed it so that we don't need to do any `std`-specific conditional compilation in files other than the `io` or `lib` facades (but couldn't make it proper `pub(crate)` due to `io::Result` leaking in the public API).

> I am making some tweaks in later commits (...)

Thanks for not blocking on this but it'd be completely fine to ask me to follow up with other tweaks related to this PR!

dtolnay:
Published in 1.0.45.

Cargo.toml
cuviper: > The current error if you do `serde_json = { default-features = false }` [...]

Shouldn't this new error be considered a breaking change?

PR #614, opened 2020-02-02
replace serialize key,value functions by serialize_entry

jmfiaschi:
The purpuse is to serialize serde_json::Map<> with other serializer like serde-xml-rs::to_writer(&mut writer, &json_value) that is not possible today. The problem is serde_json use only **map.serialize_key(k)** and **map.serialize_key(v)**. Some serializer can use only **map.serialize_entry(k, v)**. If we follow these advices :
https://docs.serde.rs/src/serde/ser/mod.rs.html#1778-1784
/// Serialize a map key.
///
/// If possible, `Serialize` implementations are encouraged to use
/// `serialize_entry` instead as it may be implemented more efficiently in
/// some formats compared to a pair of calls to `serialize_key` and
/// `serialize_value`.
and the implementation of serialize_entry : https://docs.serde.rs/src/serde/ser/mod.rs.html#1798-1826

We should use during the Map serialize use serialize_entry function instead of serialize_key/serialize_key and use them into serialize_entry.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -307,8 +307,7 @@ impl ser::Serialize for Map<String, Value> {
         use serde::ser::SerializeMap;
         let mut map = tri!(serializer.serialize_map(Some(self.len())));
         for (k, v) in self {
-            tri!(map.serialize_key(k));
-            tri!(map.serialize_value(v));
+            map.serialize_entry(k, v)?;
         }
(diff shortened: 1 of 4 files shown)


dtolnay:
Thanks, published in 1.0.46.

PR #621, opened 2020-02-13
Expose `value::ser::Serializer`

sdleffler:
This just makes `Serializer` public through a `pub use` in `serde_json::value`.


dtolnay:
Published in 1.0.48.

PR #628, opened 2020-03-07
Add value-trait feature

Licenser:
Hi,
we've [built a trait](https://docs.rs/value-trait/0.1.2/value_trait/) that allows for generalizing over different types of JSON values / DOM's.

I added the required implementations to the `serde_json::Value`, the implementation is behind the `value-trait` feature flag to not pollute the dependencies if it's not used.

Added a PR to serde-yaml as well to illustrate the use of this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ serde = { version = "1.0.100", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = { version = "0.4.3", default-features = false }
 ryu = "1.0"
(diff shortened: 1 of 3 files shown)


Licenser:
I understand the sentiment sadly it is not possible as some `From` traits need to be implemented on `Value` which can not be done in a remote crate.

dtolnay:
There is only one `From` impl in this PR, and that one does not need to be in serde_json.

Licenser:
sorry, wrong trait, `PartialEq` is the one.

PR #640, opened 2020-03-27
Implement Eq for Map, Number and Value.

### LINKED ISSUE #638: Implement Eq for Value, Number and Map
All these types currently implement `PartialEq`. However, they also fulfil the requirements for `Eq`, so we may implement that trait as well.

The `Map` type currently has a [custom implementation for `PartialEq`](https://github.com/serde-rs/json/blob/7fc289015003c39044ffeef42819b5ddf5188a5e/src/map.rs#L229-L243), which looks identical to what we would get by simply deriving `PartialEq`. If the `preserve_order` feature is enabled, the code is [identical to the `PartialEq` implementation for `IndexMap`](https://docs.rs/indexmap/1.3.2/src/indexmap/map.rs.html#2204-2219). For the case that `preserve_order` is disabled, the code simply defers to the implementation for `BTreeMap`. Both `IndexMap` and `BTreeMap` implement `Eq`.

The `Number` type has three variants, wrapping `u64`, `i64` and `f64` respectively. The former two are `Eq`, while the latter is only `PartialEq`. However, since floating-point numbers in JSON are always finite, we can assume them to be `Eq` as well, so we can implement `Eq` for `Number`.

Once `Map` and `Number` implement `Eq`, it can be derived for `Value`.

smarnach:
These types already have `PartialEq` implementations which define equivalence relations, so we can implement `Eq` as well. Fixes #638.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -233,6 +233,8 @@ impl PartialEq for Map<String, Value> {
     }
 }
 
+impl Eq for Map<String, Value> {}
+
 /// Access an element of this map. Panics if the given key is not present in the
 /// map.
 ///
--- a/src/number.rs
+++ b/src/number.rs
@@ -16,7 +16,7 @@ use serde::de::{IntoDeserializer, MapAccess};
 pub(crate) const TOKEN: &str = "$serde_json::private::Number";
 
 /// Represents a JSON number, whether integer or floating point.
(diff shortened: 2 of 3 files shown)


smarnach:
I failed to test locally with the `preserve_order` feature enabled. Fixed now.

Issue #360, opened 2017-09-17
Efficiency of bytes encoding


elliottslaughter:
Is `serde_json` integrated with `serde_bytes`? I was hoping that using `serde_bytes` would allow me to encode `Vec<u8>` somewhat intelligently (e.g. as a base64-encoded string, or similar). However, a very simple test seems to indicate that it doesn't do anything at all:

```
plain {"c":[123,124,125]}
bytes {"c":[123,124,125]}
```

This encoding results in an expansion ratio of up to 4:1 (depending on the values of the bytes encoded), rather than 4:3 in a base64 encoding. Needless to say, in some use cases this can be really significant reduction in message size.

The source of the test is below.

```
#[macro_use]
extern crate serde_derive;
extern crate serde_bytes;
extern crate serde_json;

#[derive(Serialize, Deserialize, Debug)]
pub struct MyPlain {
    c: Vec<u8>
}

#[derive(Serialize, Deserialize, Debug)]
pub struct MyBytes {
    #[serde(with = "serde_bytes")]
    c: Vec<u8>
}

fn main() {
    let plain = MyPlain { c: vec![123, 124, 125] };
    let plain_json = serde_json::to_string(&plain).unwrap();
    println!("plain {}", plain_json);

    let bytes = MyBytes { c: vec![123, 124, 125] };
    let bytes_json = serde_json::to_string(&bytes).unwrap();
    println!("bytes {}", bytes_json);
}
```

dtolnay:
The default encoding is not going to change but you can plug in base64 if you want.

```rust
#[derive(Serialize, Deserialize, Debug)]
struct MyBytes {
    #[serde(with = "base64")]
    c: Vec<u8>
}

mod base64 {
    extern crate base64;
    use serde::{Serializer, de, Deserialize, Deserializer};

    pub fn serialize<S>(bytes: &[u8], serializer: S) -> Result<S::Ok, S::Error>
        where S: Serializer
    {
        serializer.serialize_str(&base64::encode(bytes))

        // Could also use a wrapper type with a Display implementation to avoid
        // allocating the String.
        //
        // serializer.collect_str(&Base64(bytes))
    }

    pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
        where D: Deserializer<'de>
    {
        let s = <&str>::deserialize(deserializer)?;
        base64::decode(s).map_err(de::Error::custom)
    }
}
```

[...]

PR #656, opened 2020-04-23
Add support for alternate encodings of bytes types

jethrogb:
JSON does not natively support binary data. `serde_json` currently just supports integer arrays and a deserialization mode for raw strings. However, protocols using JSON specify their own mechanisms to handle binary data in JSON, and it's likely not to be what `serde_json` currently does.

How types in the serde data model are to be encoded should be specified by the (de)serializer, not the type. Suggestions such as https://github.com/serde-rs/json/issues/360#issuecomment-330095360 are therefore not appropriate. By specifying the encoding that way in the type, you can't (de)serialize that type anymore with a (de)serializer that does natively support bytes (such as CBOR).

This PR introduces the concept of alternate encodings for bytes types. When creating a JSON (de)serializer, you can specify the mode you want. The default mode is still integer arrays. This PR adds base64-encoding as an option. All changes can be enabled with features and should be backwards-compatible and zero-cost for existing users. The functionality is thoroughly documented with the `BytesMode` type.

There are a couple of TODO items but I wanted to gather feedback before continuing the work.

TODO:
* [ ] some kind of support for `Value as Deserializer`
* [ ] tests

cc @jseyfried

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ serde = { version = "1.0.100", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = { version = "0.4.3", default-features = false }
 ryu = "1.0"
-b64-ct = { version = "0.1", default-features = false }
+b64-ct = { version = "0.1", default-features = false, optional = true }
 
 [dev-dependencies]
 automod = "0.1"
@@ -77,3 +77,9 @@ raw_value = []
 # overflow the stack after deserialization has completed, including, but not
 # limited to, Display and Debug and Drop impls.
 unbounded_depth = []
+
+# Support alternate encoding modes for bytes. Available on Rust 1.40+
+bytes_mode = []
+
+# Support the Base64-encoding alternate bytes encoding mode
+base64 = ["bytes_mode", "b64-ct"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ serde = { version = "1.0.100", default-features = false }
 indexmap = { version = "1.2", optional = true }
 itoa = { version = "0.4.3", default-features = false }
 ryu = "1.0"
+b64-ct = { version = "0.1", default-features = false }
 
 [dev-dependencies]
 automod = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -428,6 +428,8 @@ macro_rules! tri {
 #[macro_use]
 mod macros;
 
+#[cfg(feature = "base64")]
+pub mod base64;
 pub mod de;
 pub mod error;
 pub mod map;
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -447,3 +449,132 @@ mod read;
...
+/// calls `deserialize_any`) can't be deserialized as bytes.
+///
+/// The default mode is `IntegerArray`, which is the only format Serde JSON
+/// used to support.
+///
+/// The `Base64` mode is enabled with the `base64` crate feature.
+#[cfg(feature = "bytes_mode")]
+#[derive(Clone, Eq, PartialEq, Debug, Hash)]
+#[non_exhaustive]
+pub enum BytesMode {
```
jethrogb, line 471: Not too happy about having this in `lib.rs`, but no existing module seems like a good place.

Cargo.toml
```diff
@@ -76,3 +77,9 @@ raw_value = []
 # overflow the stack after deserialization has completed, including, but not
 # limited to, Display and Debug and Drop impls.
 unbounded_depth = []
+
+# Support alternate encoding modes for bytes. Available on Rust 1.40+
```
jethrogb, line 81: Version requirement due to https://github.com/serde-rs/json/pull/656/files#diff-b4aea3e418ccdb71239b96952d9cddb6R470

agausmann:
> How types in the serde data model are to be encoded should be specified by the (de)serializer, not the type. Suggestions such as [#360 (comment)](https://github.com/serde-rs/json/issues/360#issuecomment-330095360) are therefore not appropriate. By specifying the encoding that way in the type, you can't (de)serialize that type anymore with a (de)serializer that does natively support bytes (such as CBOR).

I don't agree entirely; I haven't found a scenario where I have to implement a specific protocol _and_ support multiple data formats.

My focus when implementing a protocol (de)serializer is making it compatible with that specific protocol, and I am fine with creating custom `Deserialize` impls to deal with that. If I want to be able to (de)serialize some of that data in another format or send/receive it using another incompatible protocol, I will create another data model to match that other structure and convert between the two.

As an example, I've recently been implementing a [game data parser](https://github.com/agausmann/yafc/tree/842f07f1d606947db9a4d19b8d3320b376b09909/src/data) for one of my projects. I wanted to pick out the relevant bits from a dump of the game data, and so I made a parser for that [(`raw.rs`)](https://github.com/agausmann/yafc/blob/842f07f1d606947db9a4d19b8d3320b376b09909/src/data/raw.rs) which required some special logic on top of the existing `Deserializer` impl that I selected. However, those complex parsing rules and special helpers make it difficult to use as a general-purpose (de)serializable model, and I'd prefer a simpler model that is easier to use for the other modules. So, I implemented a more high-level data model [(`mod.rs`)](https://github.com/agausmann/yafc/blob/842f07f1d606947db9a4d19b8d3320b376b09909/src/data/mod.rs) that is meant to be persistent/cached, and that doesn't use any special serde rules, just (de)serializing the data using whatever the selected format thinks is the best.

jethrogb:
@dtolnay ping for review?

PR #658, opened 2020-04-26
Add value::to_raw_value

### LINKED ISSUE #654: Feature request: to_raw_value
It would be nice to have an equivalent to `to_value` for `RawValue`s, to simplify the following / skip parsing in it:

```rust
let json_string = serde_json::to_string(serializable_thing).unwrap();
RawValue::from_string(json_string).unwrap()
```

jplatte:
Resolves #654.

I adopted the `# Errors` doc section from `to_value`. I can also try coming up with a semi-realistic example of how one would use this.

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -215,6 +215,66 @@ impl RawValue {
     }
 }
 
+/// Convert a `T` into a boxed `RawValue`.
+///
+/// # Example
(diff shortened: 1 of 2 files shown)


jplatte:
@dtolnay Is this realistic enough?

PR #659, opened 2020-04-28
Implement `Hash` for `serde_json::Value`

Ten0:
It happens I need this for custom [sentry](https://sentry.io/) event aggregation, where events may contain extra context represented as `serde_json::Value`. Currently I cannot hash these, due to this missing impl.


Ten0:
Nevermind, I guess this is hard because this is not implemented for `IndexMap` and more importantly even for IndexMap.

PR #670, opened 2020-05-12
Remove type errors in de::Deserializer

Plecra:
In JSON, it's not uncommon to interpret values differently based on the structure of the data. For example, you might wish allow a user to use predefined strings as constants in a number field.

Today, this is implemented using `deserialize_any`, allowing the `Deserializer` to determine the type of the input. This works well, but prevents the `Deserialize` implementation from being compatible with other, non self-describing formats.

If a user instead tries to give a hint for self-describing formats, `serde_json` will quickly reject input that isn't compatible by raising `invalid_type` errors.

By treating the `deserialize_` calls as hints ([as serde seems to](https://serde.rs/impl-deserializer.html)), we can allow `Deserialize` implementations to use them while handling other types in their `Visitor` implementation.

There's a possibility that this will bloat the code generated by `Visitor` instantiations, but I suspect the default implementations won't make much of a dent. How might we review that impact?

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -5,7 +5,7 @@ use crate::lib::str::FromStr;
 use crate::lib::*;
 use crate::number::Number;
 use crate::read::{self, Fused, Reference};
-use serde::de::{self, Expected, Unexpected};
+use serde::de::{self, Unexpected};
 use serde::{forward_to_deserialize_any, serde_if_integer128};
 
 #[cfg(feature = "arbitrary_precision")]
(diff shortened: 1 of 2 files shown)


Plecra:
I need to make some changes to make the `arbitrary_precision` feature play nice. It shouldn't be a problem to reintroduce the `prim_number` code, but I want to hear your take.

Would it be unreasonable to reuse the `RawValue` api to implement `Number`?

```rust
impl<'de> Deserialize<'de> for Number {
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Number, D::Error> {
        RawValue::deserialize(deserializer)?.get().parse().map_err(D::Error::custom)
    }
}
```

agausmann:
> There's a possibility that this will bloat the code generated by Visitor instantiations

I had a [very similar conversation](https://github.com/serde-rs/serde/issues/1803) recently, and it definitely can increase binary sizes significantly. For example, when you are using a type that calls `deserialize_map`, if that forwards to `deserialize_any` which can potentially call many different visitor methods, that means that many more visitor methods will have to be instantiated, even if you're only going to call `visit_map` during normal operation and the other visitors return type errors. On the other hand, if the `deserialize_map` impl only calls `visit_map`, then the rest of the visitor methods are unreachable and don't need to be included in the binary.

Plecra:
I don't *really* get why, but it looks like the changes do actually increase the binary size by ~15kb per struct. That seems huge, and I imagine the reason is a little more complex than the extra visitor instantiations (I'd expect ~20 `E::invalid_type` calls, which would take a few hundred bytes at most.)

The problems with the `arbitrary_precision` feature turned out to be fairly complex to solve too, and I won't be able to finish this PR in the short term. I'll close it for now.

PR #687, opened 2020-06-28
Optimize compile time of Deserializer trait methods

dtolnay:
We can instantiate the control flow in these methods once per Read type rather than once per Visitor type. This reduces compile time of https://github.com/serde-rs/json-benchmark by 44% as measured by `cargo clean && RUSTFLAGS='-C codegen-units=1' cargo build --release --no-default-features --features parse-struct,lib-serde,all-files -Ztimings`, with 0-13% impact on performance.

#### Before

```console
   Completed serde_json v1.0.55 in 3.8s
   Completed json-benchmark v0.0.1 bin "json-benchmark" in 18.1s
```

```console
                                DOM                  STRUCT
data/canada.json         270 MB/s              620 MB/s
data/citm_catalog.json   390 MB/s              950 MB/s
data/twitter.json        270 MB/s              560 MB/s
```

#### After

```console
   Completed serde_json v1.0.55 in 3.7s
   Completed json-benchmark v0.0.1 bin "json-benchmark" in 10.1s
```

```console
                                DOM                  STRUCT
data/canada.json         250 MB/s              540 MB/s
data/citm_catalog.json   380 MB/s              850 MB/s
data/twitter.json        260 MB/s              560 MB/s
```

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -304,26 +304,24 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     where
         V: de::Visitor<'de>,
     {
-        let peek = match tri!(self.parse_whitespace()) {
-            Some(b) => b,
-            None => {
-                return Err(self.peek_error(ErrorCode::EofWhileParsingValue));
-            }
+        let erased_visitor = {
+            let erased_visitor = &DeserializeNumber {
+                visit_number: do_visit_number::<V>,
+                drop_visitor: do_drop::<V>,
+                expecting: do_expecting::<V>,
+            };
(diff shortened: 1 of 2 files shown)


Marwes:
This could be done in a safe manner though I guess there would be some additional branches unless the unreachable!() branches are made unreachable_unchecked. https://github.com/Marwes/json/tree/safe_once_dispatch

It is only a POC but I'd expect the vtable from the trait to end up the same as this unsafe impl.

Marwes:
cc #313

PR #690, opened 2020-06-30
refactor: Factor out whitespace skipping into helpers

Marwes:
Should hopefully reduce the amount of code being generated.

Only saw #687 after I did this, while conflicting due to moving some methods this should otherwise be orthogonal.


Marwes:
I did try this earlier with no improvements to compile time, but I think this is an improvement regardless https://github.com/serde-rs/json/issues/313#issuecomment-349049626

PR #702, opened 2020-08-08
Updated readme to address array and deep object access.

jrsmith17:
Even though I had used serde_json to (de)serialize arrays of flat JSON objects previously, I struggled for a little to determine how to navigate the JSON object itself. Once I figured it out, I thought it made sense to update the README with two examples that I personally found helpful so that other people wouldn't have to struggle as much as I did to learn the crate.

Happy to take edits on the commented descriptions. I just put in a good faith effort to describe it as it would make sense to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,59 @@ fn print_an_address() -> Result<()> {
 }
 ```
 
+## Navigating a hierarchical JSON structure
+
+Some times you'll need to navigate a JSON structure deserialized
+by serde_json. This example should help you conceptualize how to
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks for the PR! I believe indexing json objects and json arrays are both already shown in https://github.com/serde-rs/json/tree/d0d80783d31161c869922fad933b3623715506d2#operating-on-untyped-json-values, where it has `v["phones"][0]`.

PR #708, opened 2020-09-18
Add Map::remove_entry

jplatte:
… and simplify `Map::remove`, since I noticed it uses `#[cfg]` branching when it doesn't need to.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@nightly
       - run: cargo test
+      - run: cargo test --features map_remove_entry --tests -- --skip ui --exact
       - run: cargo test --features preserve_order --tests -- --skip ui --exact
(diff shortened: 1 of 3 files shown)


jplatte:
Okay so `BTreeMap::remove_entry` is only available since 1.45.0. Would feature-gating stuff like this that requires a higher MSRV be an option?

jplatte:
@dtolnay It's now behind a feature flag that is also tested on CI. Can you take a look?

dtolnay (changes requested):
I would prefer not to do feature stuff for this. It would be better to emulate the API if remove_entry is not directly available on the underlying type. Is that possible?

jplatte:
No, unfortunately I don't think that is possible. The entry API requires taking ownership of the key which defeats the whole point, and a raw entry API is only available for HashMap, not BTreeMap.

dtolnay:
Closing in favor of #712 which emulates the API when remove_entry is not available on the underlying type.

jplatte:
I see, so using a less efficient implementation on older versions. I like that :+1:

PR #714, opened 2020-10-02
Simplify Map::remove

jplatte:
Small remaining useful change from #708. The review didn't say anything about this commit, so I assume it was probably overlooked.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -116,10 +116,7 @@ impl Map<String, Value> {
         String: Borrow<Q>,
         Q: ?Sized + Ord + Eq + Hash,
     {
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
We changed this to the current code when IndexMap::remove was deprecated (https://github.com/bluss/indexmap/pull/99/commits/0c24d101db5660db1d75873bcdb1a659b1b3b99e). It looks like they reverted the deprecation in a later release to become drop in compatible with HashMap. I think I would lean toward keeping this as currently implemented in serde_json as a reminder that removing is somewhat nuanced for IndexMap.

jplatte:
Okay!

Issue #537, opened 2019-04-27
Document that serde_json doesn't escape forward slashes


xfix:
An user should manually escape the slash (for instance using https://github.com/serde-rs/json/issues/435#issuecomment-382108021) when putting JSON inside of `<script>` (yes, people do that), otherwise an XSS will occur. This ideally would be documented, as it is a potential security vulnerability.

andy128k:
Can we have one more formatter which also escapes forward slashes (and possibly also `<`, `>`, `&`)?

xfix:
I would argue that escaping `<`, `>` and `&` is a responsibility of an HTML escape library, I'm thinking about an use of JSON in `<script>`, where those characters are non-issue.

webmaster128:
Why would you want to escape forward slashes? They can only occur in JSON strings where they are perfectly valid JSON (and JavaScript, right?):

```json
{ "foo/bar": "spam //////////////////"}
```

But there is a different danger when doing this kind of embedding: there are unicode characters that are valid in JSON strings but not in JS strings: https://twitter.com/simon_warta/status/1255505650257362944

dtolnay:
@webmaster128: https://stackoverflow.com/a/1580682/6086311

andy128k:
To use serde_json in such scenarios, I had created alternative formatter https://github.com/xd009642/tarpaulin/blob/master/src/report/safe_json.rs 

It would be great if such formatter is available out of the box.

xfix:
@webmaster128 U+2028 and U+2029 are now allowed in JavaScript strings, however fair point in that it will only work on relatively modern web browsers (Chrome 66+, Edge 79+, Firefox 62+, Opera 53+, Safari 12+). This may or may not be okay.

andy128k:
There is nothing bad in embedding JSON into `<script>`. JSON **is** a subset of JS (since ES2019).

webmaster128:
Thanks for the answers. I continued the research a lottle bit to understand _why_ this is a problem. This [is described in this StackOverflow answer](https://stackoverflow.com/a/4748651/2013738) and the linked documents. It is not at all about JavaScript. It is about how HTML parsers detect the end of a `<script>`.  The following code is also valid JavaScript but [is documented to break](https://www.w3.org/TR/html4/appendix/notes.html#h-B.3.2) the `<script>` block:

```html
<SCRIPT type="text/javascript">
      document.write ("<EM>This won't work</EM>")
</SCRIPT>
```

[...]

PR #715, opened 2020-10-02
Simplify creation of custom formatters with different escaping

andy128k:
Related to #537

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1588,6 +1588,7 @@ impl CharEscape {
             self::FF => CharEscape::FormFeed,
             self::RR => CharEscape::CarriageReturn,
             self::QU => CharEscape::Quote,
+            self::FS => CharEscape::Solidus,
             self::BS => CharEscape::ReverseSolidus,
             self::UU => CharEscape::AsciiControl(byte),
             _ => unreachable!(),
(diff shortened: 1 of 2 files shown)


andy128k:
@dtolnay Could you give a feedback on this PR?

dtolnay (commented):
Do you have a need for this other than the HtmlSafe case, or is it intended mostly for that? (I'm not sure if the EscapeAllCharactersFormatter is a use case you've come across in practice.)

andy128k:
@dtolnay Practically only HtmlSafe is needed. Other examples are done only to have a test coverage and demonstrate flexibility.

dtolnay (commented):
In that case I would prefer not to add this API -- all it accomplishes is makes HtmlSafe *slightly* fewer lines to implement. Comparing with vs without, I don't think the difference is significant enough to motivate this change.

```rust
impl Formatter for HtmlSafeCompactFormatter {
    fn char_escape(&self, byte: u8) -> Option<CharEscape> {
        if byte == b'/' {
            Some(CharEscape::Solidus)
        } else {
            CompactFormatter.char_escape(byte)
        }
    }
}
```

```rust
impl Formatter for HtmlSafeCompactFormatter {
    fn write_string_fragment<W: ?Sized + Write>(&mut self, w: &mut W, s: &str) -> io::Result<()> {
        for (i, split) in s.split('/').enumerate() {
            if i > 0 {
                w.write_all(b"\\/")?;
            }
            CompactFormatter.write_string_fragment(w, split)?;
        }
        Ok(())
    }
}
```

PR #719, opened 2020-10-13
Add serde_json::error::Error::msg() method

gmorer:
Actually it is not possible to have the error message without the line and column number.

I had to parse a file with one object by line, the error line was always 1 but not necessary in the file.

You cant get the line and column with serde_json::error::Error::{line,column} but the error message from the `ErrorCode` is private.

You can now get only ```missing field `type` ```  instead of the whole ```missing field `type` at line 1 column 26```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -18,6 +18,11 @@ pub struct Error {
 pub type Result<T> = result::Result<T, Error>;
 
 impl Error {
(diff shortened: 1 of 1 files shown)


gmorer:
No idea, waiting for maintainers to approve or disapprove.

Issue #728, opened 2020-11-12
examples don't work


dataf3l:
here is the example:
```
[package]
name = "serde_examples"
version = "0.1.0"
authors = ["Felipe@MBP2 <auditor400@gmail.com>"]
edition = "2018"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
serde_json = "1.0"




```

code:

```
use serde::{Deserialize, Serialize};
use serde_json::Result;

#[derive(Serialize, Deserialize)]
struct Address {
    street: String,
    city: String,
}

fn print_an_address() -> Result<()> {
    // Some data structure.
    let address = Address {
        street: "[null,4,'\\'$\\'#,##0.00']".to_owned(),
        city: "London".to_owned(),
    };

    // Serialize it to a JSON string.
    let j = serde_json::to_string(&address)?;

    // Print, write to a file, or send to an HTTP server.
    println!("{}", j);

    Ok(())
}
```

somehow, I get this error:

error[E0432]: unresolved import `serde`
 --> src/main.rs:1:5
  |
1 | use serde::{Deserialize, Serialize};
  |     ^^^^^ use of undeclared type or module `serde`

error: cannot determine resolution for the derive macro `Serialize`
 --> src/main.rs:4:10
  |
4 | #[derive(Serialize, Deserialize)]
  |          ^^^^^^^^^
  |
  = note: import resolution is stuck, try simplifying macro imports

error: cannot determine resolution for the derive macro `Deserialize`
 --> src/main.rs:4:21
  |
4 | #[derive(Serialize, Deserialize)]
  |                     ^^^^^^^^^^^
  |
  = note: import resolution is stuck, try simplifying macro imports

error[E0601]: `main` function not found in crate `serde_examples`
  --> src/main.rs:1:1
   |
1  | / use serde::{Deserialize, Serialize};
2  | | use serde_json::Result;
3  | |
4  | | #[derive(Serialize, Deserialize)]
...  |
23 | |     Ok(())
24 | | }
   | |_^ consider adding a `main` function to `src/main.rs`

error[E0277]: the trait bound `Address: serde::ser::Serialize` is not satisfied
    --> src/main.rs:18:35
     |
18   |     let j = serde_json::to_string(&address)?;
     |                                   ^^^^^^^^ the trait `serde::ser::Serialize` is not implemented for `Address`
     |
[...]

PR #730, opened 2020-11-22
Update README.md

actuday6418:
The absence of an acknowledgement that Serde was used in the example led to a good amount of wasted time for me thinking the example was out dated code and this has apparently been an issue for others too. (https://github.com/serde-rs/json/issues/728)

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ A data structure can be converted to a JSON string by
 [`serde_json::to_string`][to_string]. There is also
 [`serde_json::to_vec`][to_vec] which serializes to a `Vec<u8>` and
 [`serde_json::to_writer`][to_writer] which serializes to any `io::Write`
-such as a File or a TCP stream.
+such as a File or a TCP stream. The following code uses the Serde serialization framework for Rust, documentation for which may be found [here](https://github.com/serde-rs/serde).
 
 <a href="https://play.rust-lang.org/?edition=2018&gist=3472242a08ed2ff88a944f2a2283b0ee" target="_blank">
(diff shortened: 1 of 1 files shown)


vallentin:
I feel like, if it should point to a link, that it would be better to point to https://serde.rs/derive.html. However, the fact that it uses/needs `serde` is somewhat self-explanatory from the name itself (`serde_json`) and the use declarations in the example(s).

Alternatively, maybe (also) have a comment before the import like this:

```rust
// serde = { version = "1.0", features = [ "derive" ] }
use serde::{Deserialize, Serialize};
use serde_json::Result;
```

One of the previous sections, "_Parsing JSON as strongly typed data structures_". Has the following note:

> The necessary setup for using Serde's derive macros is explained on the [Using derive](https://serde.rs/derive.html) page of the Serde site.

Maybe it would be good to have that note somewhere before all the examples.

shtukas:
New to Rust and somehow just had the same frustration. It wasn't obvious to me that I needed to add `serde` to my `Cargo.toml` file considering that the README has explicitly made the effort to mention `serde_json`

dtolnay (commented):
I think this is fine as written.

Serde is used by more people than any other crate, which means it is used by more beginners than any other crate too, but that doesn't mean it makes sense to evolve its documentation toward beginner teaching material; that material already exists elsewhere.

The code in this example shows `use serde::{Deserialize, Serialize}` so it should be clear that comes from a `serde` crate, and the documentation just a few lines above this example already discusses the necessary setup for using serde's derive macros.

PR #733, opened 2020-12-02
Allow collecting an iterator of pairs into JSON object

matklad:
This feels like an API omission if I am not missing something.


dtolnay (approved):
Thanks!

PR #735, opened 2020-12-21
[RFC] Seeded `StreamDeserializer`

marmeladema:
I am finding myself in the need of a "seeded" `StreamDeserializer` (ie: a `StreamDeserializer` backed by a `S: DeserializeSeed` instead of a `T: Deserialize`).

My use-case is that I want to use `StreamDeserializer` but for a type that only implements `DeserializeSeed`.

This PR proposes two similar approaches where the generic type `T`: the type of the generated items, is replaced by a generic type `S`: the type of the seed, which implements some form of `DeserializeSeed` and where the generated items are of type `DeserializeSeed::Value`.

First commit is the "simple" way but requires that `S` is also `Copy` in order to be able to call `DeserializeSeed::deserialize` in each call to `next()` (this is because `DeserializeSeed::deserialize` takes ownership of `self`).

Second commit is a more convoluted approach where `&mut S:  DeserializeSeed<'de>` is required instead. This allows some more fancy seed to be used, as shown in the added test.

I *think* those approaches are backward compatible.

I also added the `RFC` tag in the PR title because I would like to get some feedbacks about that idea itself and if it seems a sensible requirement, around the proposed implementations.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -136,7 +136,7 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     }
 
     /// Turn a JSON deserializer into an iterator over values of type T.
-    pub fn into_iter<T>(self) -> StreamDeserializer<'de, R, T>
+    pub fn into_iter<T>(self) -> StreamDeserializer<'de, R, PhantomData<T>>
     where
         T: de::Deserialize<'de>,
     {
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
I would prefer not to add any new API for this. For the use case in your test, something like the following should work:

```rust
let data = b"[0] [1,2] [3] []";
let mut ints = Vec::<usize>::new();

let mut de = serde_json::Deserializer::from_slice(data);
loop {
    match ExtendVec(&mut ints).deserialize(&mut de) {
        Ok(()) => {}
        Err(done) if done.is_eof() => break,
        Err(err) => return eprintln!("{}", err),
    }
}

assert_eq!(ints, &[0, 1, 2, 3]);
```

PR #737, opened 2020-12-28
Add impl From<Number> for Value

### LINKED ISSUE #736: impl From<Number> for Value is missing
This could be useful at times. The concrete use case I have is a string that needs to be converted to `Value`. The string could be either numerical in which case it should be `Value::Number` or just a text, in which case it will be `Value::String`.
So I am using next code:
```rust
// text here is &str
if let Ok(number) = text.parse::<Number>() {
    Value::Number(number)
} else {
    Value::from(text)
}
```

Writing `Value::from(number)` feels cleaner

imp:
Closes #736

### DIFF
--- a/src/value/from.rs
+++ b/src/value/from.rs
@@ -136,6 +136,22 @@ impl<'a> From<Cow<'a, str>> for Value {
     }
 }
 
+impl From<Number> for Value {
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #739, opened 2021-01-02
make a html link in the doc comments

rory:
just a minor ergonomic improvement to the documents on the web


dtolnay (approved):
Thanks!

dtolnay:
I changed this to a module system based link in 44d7fe33fcd1adfb952384367840d706b7278d9e. Otherwise https://docs.rs/serde_json/1/serde_json/enum.Value.html would point to the wrong index page compared to https://docs.rs/serde_json/1/serde_json/value/enum.Value.html.

PR #744, opened 2021-01-10
Update example to use Result<User, Box<dyn Error>>

Carreau:
This seem to be what is now expected on stable; as the compiler warnings are saying; and also consistent with what `read_user_from_stream` example is using.


dtolnay (approved):
Thanks!

Issue #750, opened 2021-02-05
PrettyFormatter::with_indent allows generating non-utf8 output


icewind1991:
Because `PrettyFormatter::with_indent` takes a `&[u8]` as indent a user can generate a non-utf8 serialized output by providing a non-utf8 indent.

This doesn't currently provide any unsoundness as currently you can only serialize using a custom `PrettyFormatter` to an `io::Write` which doesn't give any guarantees that the output is valid json. But it still makes it easy to accidentally break the "We do not emit invalid UTF-8" guaranties, or create issues downstream for any user not realizing that that guarantee doesn't always hold when serializing directly to an `io::Write`.

So while there is probably no need for direct action here, changing the ident to a `&str` whenever a new major version is released removes any possible problems here.

PR #751, opened 2021-02-05
use from_utf8_unchecked in the fmt::Display implementation of Value

icewind1991:
Since the serializer only emits valid utf8 [by default](https://github.com/serde-rs/json/issues/750) we can remove the checks that the data being formatted is valid.

`WriterFormatter` has been moved into it's own little module to enforce using it's `new` method to convey the unsafe requirements.
the safety requirements that the serializer only emits valid utf8 is already assumed by the existing `to_string` methods.

Marking `WriterFormatter` as unsafe is not strictly necesarry as it is already contained within the module bounds, but this way it's a bit more explicit about the assumptions being made.

this speeds up formatting large `Value`s by about 33% in my testing.

### DIFF
--- a/src/value/mod.rs
+++ b/src/value/mod.rs
@@ -91,7 +91,6 @@
 //! [from_reader]: https://docs.serde.rs/serde_json/de/fn.from_reader.html
 
 use crate::error::Error;
-use crate::io;
 use crate::lib::*;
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #754, opened 2021-02-25
Add Entry::and_modify

Krout0n:
Like `std::collections::hash_map::Entry`, I guess the `Entry` defined in this crate can provide `and_modify`.


dtolnay (approved):
Thanks!

PR #758, opened 2021-03-01
Refactor Value::pointer_functions by using Iterator::try_fold

Krout0n:
By using Iterator::try_fold, we don't need to go a roundabout route for borrow checker anymore.


dtolnay (approved):
Thanks! This code is from Rust 1.6, which was way before try_fold.

PR #760, opened 2021-03-09
Support streaming iteration over sequences

### LINKED ISSUE #404: Provide a way for StreamDeserializer to stream an array of values

leo60228:
Closes #404 

This makes it possible to get a `SeqAccess` outside of a `Visitor` implementation, and provides a wrapper implementing `Iterator` for convenience.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -2014,30 +2014,23 @@ pub struct SeqIntoIter<'a, R: 'a, T> {
     _phantom: PhantomData<fn() -> T>,
 }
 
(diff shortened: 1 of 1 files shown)


leo60228:
Ugh, I completely forgot that serde had such a low MSRV...

dtolnay (commented):
Thanks for the PR!

This looks generally equivalent to #653, so I would prefer not to accept it for the same reason.

PR #767, opened 2021-04-02
Updating README.md for Object that doesn't work. 

CarlyRaeJepsenStan:
Object() doesn't work - updated this for an any-typed object using HashMap.  Also, added updated code/playground links for an example that uses HashMap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,19 +139,22 @@ in one of the dozens of places it is used in your code.
 Serde provides a powerful way of mapping JSON data into Rust data structures
 largely automatically.
 
-<a href="https://play.rust-lang.org/?edition=2018&gist=15cfab66d38ff8a15a9cf1d8d897ac68" target="_blank">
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
I would prefer not to make this change. Though I don't know what you mean by *"Object() doesn't work"*.

Thanks anyway!

README.md
```diff
@@ -71,7 +71,7 @@ enum Value {
     Number(Number),
     String(String),
     Array(Vec<Value>),
-    Object(Map<String, Value>),
+    HashMap<String, Value>,
```
dtolnay, line 74: What you've put here is not valid Rust syntax.

CarlyRaeJepsenStan:
> Though I don't know what you mean by _"Object() doesn't work"_.

When I tried to run the pre-existing example for parsing objects inside of JSONs, it doesn't work - the console has no output.   Sorry it wasn't clear!  I made an edit showing how to have Serde understand objects inside of JSONs to help other people avoid the time I spent digging around.  If ` HashMap<String, Value>,` isn't correct syntax, what would be?

PR #773, opened 2021-04-15
Update link to serde-json-core

ratijas:
Seems like the project moved to a new location, and the old one returns 404.


dtolnay (approved):
Thanks!

PR #780, opened 2021-05-25
Panic on wrong call order to serialize_{key,value} and end

koivunej:
It is possible produce invalid json if one in custom `serde::Serialize` implementation forgets the `SerializeMap::serialize_value` between `SerializeMap::serialize_key` and `SerializeMap::end` as illustrated in the first commit. This was fixed by adding panics to both `src/ser.rs` and `src/value/ser.rs`:

- calling `end` when `serialize_value` was expected
- calling ` serialize_value` when `serialize_key` was expected

Previously:
- the only case for a panic with `serde_json::value` was when `serialize_value` was called before `serialize_key`
- the writer serialization (`src/ser.rs`) allowed producing `{"key"}` or `{:"value"}`

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -499,9 +499,14 @@ where
 #[doc(hidden)]
 #[derive(Eq, PartialEq)]
 pub enum State {
+    /// No content is expected, as the number of elements were previously given as zero.
     Empty,
+    /// First elements key is about to be serialized
     First,
+    /// N key value pairs have been serialized
     Rest,
+    /// A key has been serialized but not the value
+    Value,
 }
 
 // Not public API. Should be pub(crate).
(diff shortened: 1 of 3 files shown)


koivunej:
Will fix the clippy today-ish.

koivunej:
Oki, now this should now pass! I noticed that nightly clippy is used, but only with default target, where as there are 5 warnings with `--all-targets` at least initially visible. This would probably make for a nice good first issue to fix them?

koivunej:
Rebasing shortly.

koivunej:
@dtolnay pinging you because I've recently discovered github does not send emails on pending "allow workflow run" presses, and of course, any other feedback you provide. There is also another PR waiting for execution.

dtolnay (commented):
Thanks for the PR!

I don't think serde_json needs to perform this runtime check. Producing wrong output in response to a wrong Serialize impl seems fine to me.

koivunej:
This PR came out of an invalid json being produced by an existing crate under some conditions, and the checks would be quite fail fast way enabling faster learning for users.

Please ping me if you would welcome a `#[cfg(debug_assertions)]` version instead where the checks would only be done in debug builds. If not, thanks for your continued work and for taking a look at this.

PR #786, opened 2021-07-17
Make arbitrary_precision preserve the exact string representation

### LINKED ISSUE #785: With feature "arbitrary_precision" exponent marker get converted to lowercase
Comment in [`Cargo.toml`](https://github.com/serde-rs/json/blob/ea39063f9c8c7dde4ac43bbd9843287bece59b1a/Cargo.toml#L71) for feature `arbitrary_precision` states:
> This feature makes `JSON` -> `serde_json::Number` -> `JSON` produce output identical to the input.

But in [`serde_json::de::scan_exponent`](https://github.com/serde-rs/json/blob/ea39063f9c8c7dde4ac43bbd9843287bece59b1a/src/de.rs#L934), for the exponent marker (`e` or `E`), a lowercase `e` is inserted unconditionally to the internal representation. Therefore, for an input JSON `1.9E10` for example, a "`JSON` -> `serde_json::Number` -> `JSON`" roundtrip results in JSON `1.9e10`, which is different from the input JSON.

Is this behaviour intended? If no other code rely on `e` being lowercase, I suggest we insert an `e` or `E` based on the input. Some specification requires some fields always treated as string, and in case they are presented as numbers in JSON, they should be converted to strings *verbatim*. `arbitrary_precision` is the only method I found for such circumstances, and I believe it should behave as described in `Cargo.toml`.

P.S. If you see the above-mentioned change appropriate, I can make a PR for this.

ruifengx:
Fixes #785.

Now the exact representation of the input JSON is preserved in the internal representation, and thus really makes "JSON -> Number -> JSON" produce output identical to the input (as described in [Cargo.toml](https://github.com/serde-rs/json/blob/ea39063f9c8c7dde4ac43bbd9843287bece59b1a/Cargo.toml#L71)).

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -898,7 +898,7 @@ impl<'de, R: Read<'de>> Deserializer<R> {
     fn scan_number(&mut self, buf: &mut String) -> Result<()> {
         match tri!(self.peek_or_null()) {
             b'.' => self.scan_decimal(buf),
-            b'e' | b'E' => self.scan_exponent(buf),
+            c @ (b'e' | b'E') => self.scan_exponent(c as char, buf),
             _ => Ok(()),
         }
     }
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #788, opened 2021-07-27
Make use of nightly doc_cfg on docs.rs

jplatte:
I think this will make it much easier to tell at a glance that the `RawValue` API is not provided by default. While I was at it, I also added the attribute for other things where it seemed to make sense. Let me know what you think!


dtolnay (approved):
Thanks!

PR #790, opened 2021-08-18
Use `NotNan<f64>` to implement `Eq`, `Hash`, `PartialOrd` and `Ord` for `Number`, `Map` and `Value`.

timothee-haudebourg:
Float values in JSON can never be `NaN`, and thus can always be compared and hashed. I exploited this property with the [`ordered-float` crate](https://crates.io/crates/ordered-float) to implement `Eq`, `Hash`, `PartialOrd` and `Ord` for the `Number`, `Map` and `Value` types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ serde = { version = "1.0.100", default-features = false }
 indexmap = { version = "1.5", optional = true }
 itoa = { version = "0.4.3", default-features = false }
 ryu = "1.0"
+ordered-float = "2.7"
 
 [dev-dependencies]
 automod = "1.0"
--- a/src/map.rs
+++ b/src/map.rs
@@ -11,6 +11,8 @@ use crate::lib::iter::FromIterator;
 use crate::lib::*;
 use crate::value::Value;
 use serde::de;
+use std::cmp::Ordering;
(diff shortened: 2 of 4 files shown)


timothee-haudebourg:
Oh, something went wrong...

timothee-haudebourg:
I see, my bad, I haven't been very thorough with my testing.

timothee-haudebourg:
The issue is that `IndexMap` is a hash map implementation, contrarily to `BTreeSet`, so it cannot implement `Ord` nor `Hash` (well it could implement `Hash` but I think they are waiting to copy the standard `HashSet` `Hash` impl which has not yet been determined).

There is no easy solution for this, so for now I feature gated my contribution so that `Value` only implement `Ord` and `Hash` when the `preserve_ord` feature is disabled (which is kind of ironic for `Ord`). I still think it is better than nothing.

dtolnay (commented):
I don't think this will work, since cargo features are required to be additive. It can't **remove** a trait impl when a crate somewhere else in your dependency graph enables the feature.

timothee-haudebourg:
Right, I see how that could be problematic. Then the only solution would be some kind of `BTreeIndexMap` as discussed in https://github.com/bluss/indexmap/issues/199.

PR #797, opened 2021-09-12
mention `serde_json::from_value` in the doc

RouquinBlanc:
it feels wrong not to mention `serde_json::from_value` the same way as `from_slice` and `from_buffer`.


RouquinBlanc:
This is a follow up of https://stackoverflow.com/questions/69146750

I had difficulties from the general doc to find this was supported, so adding a few lines about `from_value` in there seemed like a good idea!

PR #800, opened 2021-09-14
Parse -0 as float -0.0 instead of integer 0

### LINKED ISSUE #799: Preserve sign on negative zero f32
`-0` is parsed as `0` for `f32` type.

Is it intentional? I know that this is an edge case, but this is somewhat important for my use case, because it technically two different values.

rustc 1.54.0, serde_json v1.0.67

dtolnay:
Fixes #799.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -434,8 +434,8 @@ impl<'de, R: Read<'de>> Deserializer<R> {
                 } else {
                     let neg = (significand as i64).wrapping_neg();
 
-                    // Convert into a float if we underflow.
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
Ugh, GitHub Actions hates the branch name.

```console
Checking out the ref
  /usr/bin/git checkout --progress --force -B -0 refs/remotes/origin/-0
  Error: fatal: '-0' is not a valid branch name.
  Error: The process '/usr/bin/git' failed with exit code 128
```

PR #801, opened 2021-09-14
Parse -0 as float -0.0 instead of integer 0

### LINKED ISSUE #799: Preserve sign on negative zero f32

dtolnay:
Fixes #799.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -434,8 +434,8 @@ impl<'de, R: Read<'de>> Deserializer<R> {
                 } else {
                     let neg = (significand as i64).wrapping_neg();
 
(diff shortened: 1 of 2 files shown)


silverjam:
This seems to break a "round trip" of a `0` value.  We have a test that does a conversion from JSON -> typed structure then back to JSON, when we compare the reference JSON to the generated JSON we get a breakage because `0` is not equal to `-0`.

PR #814, opened 2021-10-20
Implement `Hash` for `Number`.

timothee-haudebourg:
This is the minimal version of my previous PR #790. I implemented the `Eq` and `Hash` traits for the `Number` type, regardless of its definition. Contrarily to my previous PR, this is strictly limited to the `Number` type.

I was originally going to use the `ordered_float` crate for that, but it turns out it is much simpler to do it manually. It avoids a dependency, and this library is not compatible with Rust 1.31.0 because it uses const parameters somewhere in the code of an optional feature, which has a syntax that is not supported with this version of Rust.

I had to manually implement `PartialEq` for `Number` to please clippy.

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -54,7 +54,7 @@ impl core::hash::Hash for N {
             N::PosInt(i) => i.hash(h),
             N::NegInt(i) => i.hash(h),
             N::Float(f) => {
-                // Using `f64::to_bits` here is fine since any float values are never `Nan`.
+                // Using `f64::to_bits` here is fine since any float values are never `NaN`.
                 f.to_bits().hash(h);
             }
         }
--- a/src/number.rs
+++ b/src/number.rs
@@ -22,13 +22,25 @@ pub struct Number {
 }
 
 #[cfg(not(feature = "arbitrary_precision"))]
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone)]
 enum N {
     PosInt(u64),
     /// Always less than zero.
     NegInt(i64),
     /// Always finite.
     Float(f64),
+}
+
+#[cfg(not(feature = "arbitrary_precision"))]
+impl PartialEq for N {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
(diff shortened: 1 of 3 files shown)


src/number.rs
```diff
@@ -31,10 +31,36 @@ enum N {
...
+        }
+    }
+}
+
 // Implementing Eq is fine since any float values are always finite.
 #[cfg(not(feature = "arbitrary_precision"))]
 impl Eq for N {}
 
+#[cfg(not(feature = "arbitrary_precision"))]
+impl core::hash::Hash for N {
```
dtolnay, line 51: https://doc.rust-lang.org/1.56.0/std/hash/trait.Hash.html#hash-and-eq
`Hash` has a documented requirement that `k1 == k2` implies `hash(k1) == hash(k2)`, which this implementation violates.

```rust
use serde_json::Number;
use std::hash::{Hash, Hasher};
use std::collections::hash_map::DefaultHasher;

fn main() {
    let k1 = serde_json::from_str::<Number>("0.0").unwrap();
    let k2 = serde_json::from_str::<Number>("-0.0").unwrap();
    dbg!(k1 == k2);
    dbg!(hash(k1) == hash(k2));
}

fn hash(obj: impl Hash) -> u64 {
    let mut hasher = DefaultHasher::new();
    obj.hash(&mut hasher);
    hasher.finish()
}
```

```console
[src/main.rs:8] k1 == k2 = true
[src/main.rs:9] hash(k1) == hash(k2) = false
```

dtolnay (changes requested):
Thanks, this mostly looks good.

src/number.rs
timothee-haudebourg: Right, I forgot about that.

timothee-haudebourg:
I've added a test in the `hash` function to use the +0 hash for both +0 and -0. I've also added a test that cover this case.

With the `arbitrary_precision` feature, +0 and -0 are not equals so no problem here.

timothee-haudebourg:
Oh right, my test does not work with `arbitrary_precision` of course, let me fix that.

dtolnay (approved):
Thanks!

PR #821, opened 2021-11-11
Add `Map::get_key_value` method.

timothee-haudebourg:
I have added the `get_key_value` method to the `Map` type. This is useful to get a reference to the matched key that lives as long as the map.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -99,6 +99,7 @@ impl Map<String, Value> {
     /// The key may be any borrowed form of the map's key type, but the ordering
     /// on the borrowed form *must* match the ordering on the key type.
     #[inline]
+    #[cfg(any(feature = "preserve_order", not(no_btreemap_get_key_value)))]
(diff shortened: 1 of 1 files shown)


timothee-haudebourg:
Checks are not done yet, but I suspect it will fail with Rust below 1.40 where the `get_key_value` has been introduced in the standard library... Is backward compatibility below 1.40 a hard requirement?

timothee-haudebourg:
I see you already use a configuration predicate called `no_btreemap_get_key_value` to use the `get_key_value` in `Map::remove_entry`. That would allow me to implement the `get_key_value` method only when it is available in the inner implementation, but I don't know anything about `no_btreemap_get_key_value`. Where is it defined?

timothee-haudebourg:
So it works using `no_btreemap_get_key_value`, even though I don't know where this predicate is defined.

dtolnay (approved):
Thanks!

PR #822, opened 2021-11-12
Add retain to Map

deankarn:
This adds the `retain` function to the Map type which calls the inner
maps, Either `indexmap` or std `BTreeMap`'s, retain.

I'm trying to mutate a `Value::Object`'s Map in-place and could't find a
nice way without the `retain` function.

These new `retain` functions will also only be included when compiling with Rust 1.53.0+ when `retain` was added to the BTreeMap in the std lib https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#stabilized-apis

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -31,6 +31,12 @@ fn main() {
     if minor < 45 {
         println!("cargo:rustc-cfg=no_btreemap_remove_entry");
     }
+
+    // BTreeMap::retain
+    // https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#library-changes
+    if minor < 53 {
+        println!("cargo:rustc-cfg=no_btreemap_retain");
+    }
 }
 
 fn rustc_minor_version() -> Option<u32> {
--- a/src/map.rs
+++ b/src/map.rs
@@ -247,7 +247,7 @@ impl Map<String, Value> {
         self.map.retain(f);
     }
 
-    #[cfg(not(feature = "preserve_order"))]
+    #[cfg(all(not(feature = "preserve_order"), not(no_btreemap_retain)))]
     /// Retains only the elements specified by the predicate.
     ///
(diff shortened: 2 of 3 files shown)


build.rs
```diff
@@ -31,6 +31,12 @@ fn main() {
     if minor < 45 {
         println!("cargo:rustc-cfg=no_btreemap_remove_entry");
     }
+
+    // BTreeMap::retain
+    // https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#library-changes
```
dtolnay, line 36: Broken link

```suggestion
    // https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html#stabilized-apis
```

src/map.rs
```diff
@@ -234,6 +234,32 @@ impl Map<String, Value> {
...
+    ///
+    /// In other words, remove all pairs `(k, v)` such that `f(&k, &mut v)` returns `false`.
+    /// The elements are visited in ascending key order.
+    #[inline]
+    pub fn retain<F>(&mut self, f: F)
+    where
+        F: FnMut(&String, &mut Value) -> bool,
+    {
+        self.map.retain(f);
+    }
```
dtolnay, line 261: Why are there two identical copies of this function? (Maybe I'm not observant enough to notice the difference &mdash; what is the difference between these?)

dtolnay (changes requested):
Thanks!

src/map.rs
deankarn: Thanks for catching my mistake @dtolnay 

They were originally different but eventually became the same. Combination of late night coding and distraction of kids. It should be good to go now.

dtolnay (approved):
Looks good!

dtolnay:
Published in 1.0.70.

PR #824, opened 2021-11-13
Add impl From<Box<RawValue>> for Box<str>

jplatte:
Similar to the std `impl From<Box<str>> for Box<[u8]>`:
https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-From%3CBox%3Cstr%2C%20A%3E%3E

Use case: https://github.com/tokio-rs/axum/pull/509

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -216,6 +216,12 @@ impl RawValue {
     }
 }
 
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

Issue #335, opened 2017-06-19
serde_json::Error ought to derive a few more traits


generalelectrix:
At the very least, it would be great if serde_json::Error derived Clone.  Eq and PartialEq would be nice bonuses.  This would make the library more ergonomic and improve testability as it allows a direct comparison of expected errors.  Clone would be nice if a serialization error needs to get passed to more than one thread across a channel.

generalelectrix:
Hmm, well, it looks like even the std library error types generally don't provide Clone.  I'll just close this and go write newtype wrappers.

brycefisher:
I'd like to request that this issue be re-opened. Without deriving `PartialEq` its impossible to test using `asserrt_eq!`. What's the best way to test code that uses `serde_json` internally?

jplatte:
@dtolnay I'd also like this to be reconsidered. In ruma-events, I currently have to `.to_string()` json errors because I want structs that indirectly contain them to implement `Clone`.

tyranron:
@dtolnay

+1 for the case. Lack of `Clone` implementation pushes developers to use `.to_string()` ending up with stringly-typed opaque errors.

qwandor:
+1 for this, I'll send a PR.

legoktm:
I stumbled upon this while trying to see if someone else had already asked for `Clone` on the Error type. https://github.com/seanmonstar/reqwest/issues/1053#issuecomment-705715212 suggested wrapping non-Clone errors in an `Arc`, which fulfilled my use case, I hope that might be helpful to others.

PR #825, opened 2021-11-16
Implement Eq and PartialEq for Error.

qwandor:
`std::io::Error` values are compared by `ErrorKind`, or debug string if it is `Other`.

This partially implements #335. `Clone` is more difficult because `std::io::Error` is not `Clone`, and there is no obvious way to make an identical instance in general. It would be necessary to wrap it in an `Arc`, but that's not possible in the no-std case.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,6 +7,7 @@ use serde::{de, ser};
 
 /// This type represents all possible errors that can occur when serializing or
 /// deserializing JSON data.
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
I would prefer not to put these impls into serde_json. It should be clear that `format!("{:?}", l) == format!("{:?}", r)` is a hack &mdash; downstream code can already do the same thing if this is the behavior that it wants.

Issue #827, opened 2021-11-23
Provide an option to replace lone surrogates in strings with replacement characters


lucacasonato:
This is a continuation of #495.

JavaScript engines unilaterally agree that lone surrogates are valid inside of JSON strings (V8: https://bugs.chromium.org/p/v8/issues/detail?id=11193, SpiderMonkey: https://bugzilla.mozilla.org/show_bug.cgi?id=1496747, JSC: try `JSON.stringify('\u{1f3b5}'[0])` in the Safari console). All of their JSON parsers accept lone surrogates (try `JSON.stringify(JSON.parse(JSON.stringify('\u{1f3b5}'[0])))` in their relevant DevTools consoles).

`serde_json` always errors on lone surrogates right now. This is understandable in the context of JS representing it's strings as WTF-16, while in Rust they are always valid UTF-8. The unfortunate reality is that JS engines produce JSON text with these lone surrogates though, and these JSON messages need to be parsed sometimes. As `serde_json` aims to bridge the gap between JSON and Rust, it would be great if `serde_json` provided a way to decode strings with lone surrogates, where the surrogates are replaced by the replacement char instead of erroring.

What about something along the lines of `serde_json::from_str_lossy`?

dtolnay:
Those inputs are already supported by deserializing to byte strings. You can't directly put the deserialization result into a String because of the utf-8 invariant, but you can deserialize to byte strings and perform the lossy conversion yourself if you need it.

```rust
fn main() {
    let bytes = [b'"', "\u{1f3b5}".as_bytes()[0], b'"'];
    assert!(serde_json::from_slice::<String>(&bytes).is_err());
    let v = serde_json::from_slice::<serde_bytes::ByteBuf>(&bytes).unwrap();
    println!("{:?}", v);
}
```

lucacasonato:
Oh that's very interesting - I was not aware strings could be deserialized into byte buffers. Unfortunately I don't think this solves the problem though, as the embedder now needs to handle dealing with escape sequences manually (essentially maintaining a copy of the code from `serde_json`) that does this.

Your example is also not entirely accurate to what a JS engine will emit: it emits lone surrogates and other malformed code points as escape sequences in the JSON, rather than as the code points themselves (https://github.com/tc39/proposal-well-formed-stringify). For example: `JSON.stringify('\u{1f3b5}'[0])` will return the string `"\ud83c"`. A more accurate Rust example would be. This example showcases the issue well:

```rs
fn main() {
[...]

PR #828, opened 2021-11-24
Deserialize lone surrogates into byte bufs

lucacasonato:
This commit deserializes lone surrogates in strings that are encoded in
escape sequences instead of erroring on them.

As per https://github.com/serde-rs/json/issues/827#issuecomment-976983620

---

Note the implementation here is ugly & unoptimized. I'll go fix this after
an initial review of the functionality.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1581,18 +1581,18 @@ impl<'de, 'a, R: Read<'de>> de::Deserializer<'de> for &'a mut Deserializer<R> {
     ///
     /// Backslash escape sequences like `\n` are still interpreted and required
     /// to be valid. `\u` escape sequences are required to represent valid
-    /// Unicode code points, except in the case of lone surrogate pairs.
+    /// Unicode code points, except in the case of lone surrogates.
     ///
     /// ```
     /// use serde_bytes::ByteBuf;
     ///
     /// fn look_at_bytes() {
-    ///     let json_data = b"\"lone surrogate pair: \\uD801\"";
+    ///     let json_data = b"\"lone surrogate: \\uD801\"";
     ///     let parsed: Result<ByteBuf, _> = serde_json::from_slice(json_data);
     ///
     ///     assert!(parsed.is_ok());
     ///
-    ///     let expected = b"lone surrogate pair: \xED\xA0\x81";
+    ///     let expected = b"lone surrogate: \xED\xA0\x81";
     ///     let bytes: ByteBuf = parsed.unwrap();
     ///     assert_eq!(expected, &bytes[..]);
     /// }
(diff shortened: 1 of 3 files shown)


src/de.rs
```diff
@@ -1580,20 +1580,21 @@ impl<'de, 'a, R: Read<'de>> de::Deserializer<'de> for &'a mut Deserializer<R> {
     /// ```
     ///
     /// Backslash escape sequences like `\n` are still interpreted and required
-    /// to be valid, and `\u` escape sequences are required to represent valid
-    /// Unicode code points.
+    /// to be valid. `\u` escape sequences are required to represent valid
+    /// Unicode code points, except in the case of lone surrogate pairs.
```
dtolnay, line 1584: "lone surrogate pair" is a contradiction. Surrogate pair refers to 2 surrogates and lone surrogate refers to 1 surrogate.

src/read.rs
```diff
@@ -851,13 +865,44 @@ fn parse_escape<'de, R: Read<'de>>(read: &mut R, scratch: &mut Vec<u8>) -> Resul
...
                 // two hex escapes, representing UTF-16 surrogates.
+                // If `validate` is false and we only find a single
+                // hex escape that is a surrogate, then we'll accept
+                // it instead of erroring.
                 n1 @ 0xD800..=0xDBFF => {
-                    if tri!(next_or_eof(read)) != b'\\' {
-                        return error(read, ErrorCode::UnexpectedEndOfHexEscape);
+                    if tri!(peek_or_eof(read)) != b'\\' {
+                        if validate {
+                            tri!(next_or_eof(read));
```
dtolnay, line 874:
```suggestion
                            read.discard();
```

src/read.rs
```diff
@@ -851,13 +865,44 @@ fn parse_escape<'de, R: Read<'de>>(read: &mut R, scratch: &mut Vec<u8>) -> Resul
...
+
+                        let utf8_bytes = [
+                            (n1 >> 12 & 0x0F) as u8 | 0b1110_0000,
+                            (n1 >> 6 & 0x3F) as u8 | 0b1000_0000,
+                            (n1 & 0x3F) as u8 | 0b1000_0000,
+                        ];
+
+                        scratch.extend_from_slice(&utf8_bytes);
+
+                        return Ok(());
```
dtolnay, line 903: This doesn't have reasonable behavior for a surrogate followed by non-`\u` escape. For example `"\ud83c\n"` would decode to [237, 160, 188, 'n'] which is pretty unexpected since there is no `n` in the input.

src/read.rs
```diff
@@ -851,13 +865,44 @@ fn parse_escape<'de, R: Read<'de>>(read: &mut R, scratch: &mut Vec<u8>) -> Resul
```
dtolnay, line 865: The documentation on deserialize_bytes says it permits lone surrogates, but this codepath is still returning error on a 0xDC00..=0xDFFF surrogate.

dtolnay (changes requested):
Thanks!

lucacasonato (commented):
I have addressed your review comments, and have added some test cases for the "non \u escape code after lone surrogate" cases.

dtolnay (approved):
Looks good &mdash; thanks.

PR #841, opened 2022-01-01
Allow creating `RawValue`s from references to unsized values

EFanZh:
Currently, we cannot creating `RawValue` from `&str` using `serde_json::value::to_raw_value` since the function requires `Sized` implicitly. This PR removes this restriction.


dtolnay (approved):
Thanks!

PR #851, opened 2022-01-22
Support deserializing map key as &RawValue

dtolnay:


jplatte:
This also affects `Box<RawValue>`, correct?

Issue #855, opened 2022-01-25
Deserializing a flattened struct containing a `ByteBuf` fails to deserialize with UTF-8.


scrblue:
As per the comment for `deserialize_bytes`, it is expected that you can deserialize a non-UTF-8 string into a `ByteBuf` without failure. However, with flattened structures, `deserialize_map` is called in place of `deserialize_struct` (see https://github.com/serde-rs/serde/issues/1529), meaning members are deserialized with `deserialize_any` bypassing `deserialize_bytes`. `deserialize_any` assumes values surrounded by quotation marks are valid UTF-8 strings and returns an error otherwise.

An example follows
```rs
use serde::{Deserialize, Serialize};

fn main() {
    let a_success: A = serde_json::from_slice(b"{\"b\": {\"buf\": \"\xe5\x00\xe5\"}}").unwrap();
    println!("A Success: {:?}", a_success);

    let a_flat_success: AFlat = serde_json::from_slice(b"{\"buf\": \"abc\"}").unwrap();
    println!("A Flat Success: {:?}", a_flat_success);
    
    let a_flat_failure: AFlat = serde_json::from_slice(b"{\"but\": \"\xe5\x00\xe5\"}").unwrap();
}

#[derive(Debug, Deserialize, Serialize)]
struct A {
    b: B,
}

#[derive(Debug, Deserialize, Serialize)]
struct AFlat {
    #[serde(flatten)]
    b: B,
}

#[derive(Debug, Deserialize, Serialize)]
struct B {
    buf: serde_bytes::ByteBuf,
}
```

scrblue:
As reported in the issue I just mentioned this one in, `deserialize_any` is also used on all `enum`s with  `tag` directives meaning this deviation from expected results appears in multiple locations.

The fix you can make here would just be to ensure the `deserialize_any` function matches the functionality of  the other `deserialize_...`  functions. My PR to this crate will fix it in the case of deserializing into `ByteBuf`s, but I'm not sure if other areas are affected as no other eccentricities have appeared in my use of `serde` and `serde_json`.

To fix this in Serde would likely require more fundamental changes that I don't have the expertise to make a PR for.

jonasbb:
Since the issue mentions flattening and `deserialize_any`, the most likely culprit here is https://github.com/serde-rs/serde/issues/1183.

lucacasonato:
Also relevant is #890

PR #856, opened 2022-01-25
Allow deserializing invalid UTF-8 into bytes in maps

scrblue:
Simple fix for https://github.com/serde-rs/json/issues/855.


scrblue:
Seems like I forgot to check the tests with all features, sorry. In any case the test fails because it's expecting an error when deserializing into a `Value` where now bytes are supported in place of Strings when the data isn't valid UTF-8.

What functionality would be correct here? I don't think supporting non-UTF-8 strings is a bad choice (although I'm biased), especially since they're supported explicitly elsewhere, but I'm unsure if anyone deserializing directly into a `Value` depends on non-UTF-8 throwing an error.

Issue #807, opened 2021-09-29
Remove does not preserve order


d0cd:
The `preserve_order` feature uses an IndexMap as the backing data store for `serde_json::Map`
However, the `remove` method for IndexMap is equivalent to `swap_remove`. 
If the `preserve_order` feature is enabled, then `shift_remove` should be used instead.

dtolnay:
Closing per https://github.com/serde-rs/json/pull/860#pullrequestreview-889061502 -- I think this is all right as currently implemented.

PR #860, opened 2022-02-21
[Bug fix] Preserve order for remove operation

hanabi1224:
Changes in this PR:
- Fixed the bug described in https://github.com/serde-rs/json/issues/807
- Added unit tests

Please note that `shift_remove` is expensive. Ideally, I think we should replace [indexmap](https://crates.io/search?q=indexmap) with [hashlink](https://crates.io/crates/hashlink) which uses doubly-linked lists as internal data structure and can perform `remove`(`shift_remove`) operation much more efficiently.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -138,7 +138,7 @@ impl Map<String, Value> {
         Q: ?Sized + Ord + Eq + Hash,
     {
         #[cfg(feature = "preserve_order")]
-        return self.map.swap_remove(key);
+        return self.map.shift_remove(key);
         #[cfg(not(feature = "preserve_order"))]
         return self.map.remove(key);
     }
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
I think this is all right as currently implemented. "Preserve order" specifically refers to that you can deserialize into a `Value` and serialize the same value back without the order being scrambled. It doesn't mean arbitrary mutations aren't able to reorder keys.

If you need the more expensive `shift_remove` behavior, here is one possible way:

```rust
map.into_iter()
    .filter(|(k, _v)| k != "...")
    .collect()
```

hanabi1224:
@dtolnay Thanks for your reply. If performance is the main concern, how about switching to hashlink crate?

dtolnay:
Hashlink doesn't strike me as having a confidence-inspiring maintenance story. I don't think switching would be a good outcome.

PR #865, opened 2022-03-03
Adding serde_json::Error::code() functionnality

d34db4b3:
Hi,
With this commit, I'm trying to add the ability to identify the precise error that occurred during de/serialization. There should be no breaking change and a new method serde_json::Error::Code() now exposes the internal code of the error.

The original purpose is to provide translated information on the error (for API body request validation for example) as well as handling errors more precisely.

It is not yet perfect as serde injects some data (eg. [https://github.com/serde-rs/serde](https://github.com/serde-rs/serde/blob/7e19ae8c9486a3bbbe51f1befb05edee94c454f9/serde/src/de/impls.rs#L1174-L1176))  and changing this behavior would probably be breaking.

A simple example of the new functionality would be :
```rs
match err.code() {
    ErrorCode::InvalidType(unexp, exp) => {
        println!("Type invalide: `{}` au lieu de `{}`", unexp, exp);
    }
    ErrorCode::InvalidValue(unexp, exp) => {
        println!("Valeur invalide: `{}` au lieu de `{}`", unexp, exp);
    }
    ErrorCode::InvalidLength(len, exp) => {
        println!("Taille invalide: `{}` au lieu de `{}`", len, exp);
    }
    e => {
        println!("Other error: {}", e);
    }
}
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -42,6 +42,14 @@ impl Error {
         self.err.column
     }
 
+    /// Specifies the cause of this error.
+    ///
+    /// Usefull when precise error handling is required or translation of
+    /// error messages is required.
+    pub fn code(&self) -> &ErrorCode {
+        &self.err.code
+    }
+
     /// Categorizes the cause of this error.
     ///
     /// - `Category::Io` - failure to read or write bytes on an IO stream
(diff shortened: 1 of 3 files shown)


d34db4b3:
I've taken the errors into account and will release a fix soon.

d34db4b3:
This should be fixed now.

d34db4b3:
Fixed now 👍

d34db4b3 (commented):
This should be fixed now.
EDIT: nevermind there are still a bug with older version of rust

d34db4b3 (commented):
Fixed now 👍

d34db4b3:
Any update on the feature ?

doylemark:
Could this PR be considered for review? It seems quite important to be able to extract some information about the specific error that occurs during deserialization.

To give an example, for deserialising POST request bodies, when deserialization fails it would be nice to be able to provide meaningful errors in a 400 response explaining why the body is not formed correctly. 

Doing validation (regex/min lengths etc.) after a successful deserialization is possible, but when deserialization fails it's quite difficult to build an appropriate error message from just the `.to_string()` output and line/col number.

franksouza183:
This would be extremely useful

d34db4b3:
> This would be really great to finally get it merged. I just made this horrendous thing today:
> 
> ```rust
> if err.to_string().starts_with("missing field") {
>     // Ugly, but waiting for https://github.com/serde-rs/json/pull/865 💀
>     let parse = err.to_string().split("`");
>     let _ = parse.next();
> 
>     Err(crate::Error::MissingVariableError(
>         layer.id().to_string(),
>         parse.next(),
>     ))
> }
> ```
> 
> If the merge conflicts are the issue, I'm willing to create a new PR with them fixed

If this gets approved functionally, I can update the PR.

PR #868, opened 2022-03-13
Implement PartialEq for Value <=> &Value

### LINKED ISSUE #867: `PartialEq<&Value> for Value`?
First of all, thanks a ton for both `serde` and `serde_json`!

This is a tiny thing, but I noticed that it's not implemented.

Consider:

```rust
    assert_eq!(
        fields,
        json!({ "username": "fakeuser", "password": "fakepass" })
    );
```

...where `fields` is a `&Value`. This produces the following error:

```console
error[E0277]: can't compare `&Value` with `Value`
  --> tests/test_kbs2_new.rs:28:5
   |
28 | /     assert_eq!(
29 | |         fields,
30 | |         json!({ "username": "fakeuser", "password": "fakepass" })
31 | |     );
   | |_____^ no implementation for `&Value == Value`
   |
   = help: the trait `PartialEq<Value>` is not implemented for `&Value`
   = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Should be a small fix; I'm happy to make it if desired!

woodruffw:
Went ahead and pushed my change here, since it's small. Happy to close if it's not desired.

Closes #867.

### DIFF
--- a/src/value/partial_eq.rs
+++ b/src/value/partial_eq.rs
@@ -57,6 +57,18 @@ impl PartialEq<Value> for String {
     }
 }
 
+impl<'a> PartialEq<&'a Value> for Value {
+    fn eq(&self, other: &&Value) -> bool {
+        self.eq(*other)
+    }
+}
+
+impl<'a> PartialEq<Value> for &'a Value {
+    fn eq(&self, other: &Value) -> bool {
+        (*self).eq(other)
+    }
+}
+
 macro_rules! partialeq_numeric {
     ($($eq:ident [$($ty:ty)*])*) => {
         $($(


woodruffw:
Gentle ping for review. Let me know if there's anything else I can do!

dtolnay (commented):
From skimming https://doc.rust-lang.org/1.59.0/std/cmp/trait.PartialEq.html#implementors, the standard library has not needed this kind of impl, so I would prefer not to add it here. For example there is no PartialEq\<&String\> for String, PartialEq\<&IpAddr\> for IpAddr, PartialEq\<&i32\> for i32, etc. I am skeptical that PartialEq\<&Value\> for Value would be warranted if none of those are.

In any case, the assertion you showed in #867 would be fixed by https://github.com/rust-lang/rust/issues/44762, so I think it will eventually work without needing these extra impls.

woodruffw:
Sounds good, thanks for the explanation.

PR #869, opened 2022-03-15
allow parsing empty array into unit

jsen-:
While working on JS <-> rust interop crate, I'm trying to serialize function arguments on the javascript side and deserialize into a tuple with appropriate number of elements on the rust side. Unfortunately I found that it doesn't work for functions with no arguments.

```rust
serde_json::from_str::<(u8,u8)>("[1, 2]").unwrap(); // is fine
serde_json::from_str::<(u8,)>("[1]").unwrap(); // is fine
serde_json::from_str::<()>("[]").unwrap(); // currently fails
```
This PR allows the last line to pass.


andogq:
I would also be interested in this feature being merged! I've run into the exact same use-case. Happy to assist with any contributions required to get this merged! :)

PR #877, opened 2022-04-12
Deserialize invalid UTF-8 into byte bufs as WTF-8

lucacasonato:
Previously #828 added support for deserializing lone leading and
trailing surrogates into WTF-8 encoded bytes when deserializing a string
as bytes. This commit extends this to cover the case of a leading
surrogate followed by code units that are not trailing surrogates. This
allows for deserialization of "\ud83c\ud83c" (two leading surrogates),
or  "\ud83c\u0061" (a leading surrogate followed by "a").

The docs also now make it clear that we are serializing the invalid code
points as WTF-8. This reference to WTF-8 signals to the user that they
can use a WTF-8 parser on the bytes to construct a valid UTF-8 string.

Follow up to https://github.com/serde-rs/json/pull/830#pullrequestreview-880820431.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -861,20 +861,33 @@ fn parse_escape<'de, R: Read<'de>>(
         b'r' => scratch.push(b'\r'),
         b't' => scratch.push(b'\t'),
         b'u' => {
-            fn encode_surrogate(scratch: &mut Vec<u8>, n: u16) {
-                scratch.extend_from_slice(&[
-                    (n >> 12 & 0b0000_1111) as u8 | 0b1110_0000,
-                    (n >> 6 & 0b0011_1111) as u8 | 0b1000_0000,
-                    (n & 0b0011_1111) as u8 | 0b1000_0000,
-                ]);
+            fn encode_wtf8(scratch: &mut Vec<u8>, cp: u16) {
+                match cp {
+                    0x0000..=0x007F => {
+                        scratch.extend_from_slice(&[cp as u8]);
+                    }
+                    0x0080..=0x07FF => {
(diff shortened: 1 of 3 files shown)


src/read.rs
```diff
@@ -861,20 +861,33 @@ fn parse_escape<'de, R: Read<'de>>(
...
+                    }
+                }
             }
 
             let c = match tri!(read.decode_hex_escape()) {
                 n @ 0xDC00..=0xDFFF => {
                     return if validate {
+                        // TODO: the error message is wrong, this is a lone
+                        // _trailing_ surrogate
                         error(read, ErrorCode::LoneLeadingSurrogateInHexEscape)
```
lucacasonato, line 888: Worth adding another error code? Would doing so even be semver compatible?

tests/test.rs
```diff
@@ -1726,23 +1727,49 @@ fn test_byte_buf_de_lone_surrogate() {
...
+    assert_eq!(v, bytes);
+
+    // leading surrogate followed by U+FFFF
+    let bytes = ByteBuf::from(vec![237, 160, 188, 239, 191, 191]);
+    let v: ByteBuf = from_str(r#""\ud83c\uffff""#).unwrap();
+    assert_eq!(v, bytes);
+}
+
+#[test]
+fn test_byte_buf_de_surrogate_pair() {
```
lucacasonato, line 1763: There was no test for parsing valid surrogate pairs into byte bufs that I could find, so I added one.

lucacasonato:
@dtolnay Have you had a chance to look into this? It'd be great to get your review on this.

PR #885, opened 2022-05-03
Explicitly enable indexmap/std

cuviper:
`indexmap`'s use of `autocfg` doesn't always properly detect `std`, which is needed if you're using the default `S` hasher. The "indexmap/std" feature will explicitly force it on.

This does mean that `preserve_order` requires `std`, but that was already effectively true by using the default `S`. Alternatively, we could pick a `no_std` hasher crate for this, or implement a simple one locally.


dtolnay (approved):
Thanks!

PR #898, opened 2022-06-17
Remove solidus/slash from CharEscape.

anp:
In trying to understand how this crate does its escaping, I noticed that the enum variant and associated code exist for `CharEscape::Solidus`, but don't seem to ever be constructed/executed. Tests pass locally so I suspect this is indeed dead code, but I might be missing something.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1557,8 +1557,6 @@ pub enum CharEscape {
     Quote,
     /// An escaped reverse solidus `\`
     ReverseSolidus,
-    /// An escaped solidus `/`
-    Solidus,
     /// An escaped backspace character (usually escaped as `\b`)
     Backspace,
     /// An escaped form feed character (usually escaped as `\f`)
@@ -1777,7 +1775,6 @@ pub trait Formatter {
         let s = match char_escape {
             Quote => b"\\\"",
             ReverseSolidus => b"\\\\",
-            Solidus => b"\\/",
             Backspace => b"\\b",
             FormFeed => b"\\f",
             LineFeed => b"\\n",


dtolnay (commented):
This is for use cases that involve embedding JSON data into HTML inside &lt;script&gt;. See https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped.

```rust
use serde::Serialize;
use serde_json::json;
use serde_json::ser::CharEscape;
use std::io::{self, Write};

struct EscapeSolidus<Formatter>(Formatter);

impl<Formatter> serde_json::ser::Formatter for EscapeSolidus<Formatter>
where
    Formatter: serde_json::ser::Formatter,
{
    fn write_string_fragment<W>(&mut self, writer: &mut W, mut fragment: &str) -> io::Result<()>
    where
        W: ?Sized + Write,
    {
        if let Some(rest) = fragment.strip_prefix('/') {
            self.0.write_char_escape(writer, CharEscape::Solidus)?;
            fragment = rest;
        }
        while let Some(i) = fragment.find("</") {
            self.0.write_string_fragment(writer, &fragment[..i + 1])?;
            self.0.write_char_escape(writer, CharEscape::Solidus)?;
            fragment = &fragment[i + 2..];
        }
        self.0.write_string_fragment(writer, fragment)
    }
}

fn main() {
    let stuff = json!({
        "cmd": "head -c10 </dev/zero",
        "xss": "</script><script src=\"http://rlu.ru/344zY\"></script><script>",
    });

    let mut json = Vec::new();
    let formatter = EscapeSolidus(serde_json::ser::CompactFormatter);
    let mut serializer = serde_json::Serializer::with_formatter(&mut json, formatter);
    stuff.serialize(&mut serializer).unwrap();

    let mut out = io::stdout();
    out.write_all(b"<script>\nvar data = ").unwrap();
    out.write_all(&json).unwrap();
    out.write_all(b";\n</script>\n").unwrap();
}
```

```console
<script>
var data = {"cmd":"head -c10 <\/dev/zero","xss":"<\/script><script src=\"http://rlu.ru/344zY\"><\/script><script>"};
</script>
```

PR #900, opened 2022-06-24
Make Value be From<Option<T>>

kvnvelasco:
Adds an implementation for `From<Option<T>>` for all `T` where `T: Into<Value>`

Use case: Ergonomics for manually constructing values for insertion into `BTreeMap` (sentry does this)

I have no idea where / how the tests happen. If this is an acceptable change, could someone point me in the direction of the tests?

### DIFF
--- a/src/value/from.rs
+++ b/src/value/from.rs
@@ -275,7 +275,7 @@ where
 {
     fn from(opt: Option<T>) -> Self {
         match opt {
-            None => Self::Null,
+            None => Value::Null,
             Some(value) => Into::into(value),
         }
     }
--- a/src/value/from.rs
+++ b/src/value/from.rs
@@ -268,3 +268,15 @@ impl From<()> for Value {
         Value::Null
     }
 }
+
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Could you share how this would improve sentry's use case? What would they be writing without and with this impl?

kvnvelasco:
So sentry has a context type you can attach to errors `BTreeMap<String, Value>`. 

Almost every instance where context needs to be added to a sentry event, the value is derived from other pieces of state. Most of the time they are `Option<T>`

```rust
context.insert(
  "user_meta".into(), 
   match user_meta { 
      Some(u) => u.into(),
      None => Value::Null
   }
)
```

Is a pretty common expression. We don't really want to have to bring in a macro everywhere (because there are probably hundreds of these). 

All other non option types we send take advantage of the From implementations.

dtolnay (approved):
Thank you for the clarification. This looks good to me.

PR #904, opened 2022-07-05
Add: Copy trait to Number.

yassun4dev:
Number struct should have Copy trait because a number is literal.


dtolnay (commented):
I don't see a Copy impl being added anywhere in the code in this PR.

PR #912, opened 2022-07-19
feature: `remove_by_pointer` method

e-rhodes:
Added `Value::remove_by_pointer` method that allows you to remove a value from a nested JSON object specified by JSON pointer as in `Value::pointer{_mut}`.


angrytoastt:
This is great, thank you for showing me how to do this, I would love them to add an update and insert as well

PR #918, opened 2022-08-21
Reduce unneeded parens and newlines in Debug for Value

dtolnay:
**Before:**

```rust
println!("{:#?}", value);
```

```rust
Object({
    "Array": Array([
        Bool(
            true,
        ),
    ]),
    "Bool": Bool(
        true,
    ),
    "EmptyArray": Array([]),
    "EmptyObject": Object({}),
    "Null": Null,
    "Number": Number(
        1,
    ),
    "String": String(
        "...",
    ),
})
```

**After:**

```rust
Object {
    "Array": Array [
        Bool(true),
    ],
    "Bool": Bool(true),
    "EmptyArray": Array [],
    "EmptyObject": Object {},
    "Null": Null,
    "Number": Number(1),
    "String": String("..."),
}
```

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -307,27 +307,15 @@ impl Display for Number {
 impl Debug for Number {
     #[cfg(not(feature = "arbitrary_precision"))]
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-        let mut debug = formatter.debug_tuple("Number");
         match self.n {
-            N::PosInt(i) => {
-                debug.field(&i);
-            }
-            N::NegInt(i) => {
-                debug.field(&i);
-            }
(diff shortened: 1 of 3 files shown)


src/number.rs
```diff
@@ -307,27 +307,15 @@ impl Display for Number {
...
-                debug.field(&i);
-            }
-            N::NegInt(i) => {
-                debug.field(&i);
-            }
-            N::Float(f) => {
-                debug.field(&f);
-            }
+            N::PosInt(_) | N::NegInt(_) => write!(formatter, "Number({})", self),
+            N::Float(f) => write!(formatter, "Number({:?})", f),
```
dtolnay, line 312: Cleaned up in #919.

PR #919, opened 2022-08-21
Preserve '.0' when Displaying Number

dtolnay:
Tiny followup to #918.

```rust
let number: serde_json::Number = serde_json::from_str("1.0").unwrap();
println!("{}", number);
```

Previously: `1`
Now: `1.0`

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -292,10 +292,9 @@ impl Display for Number {
     #[cfg(not(feature = "arbitrary_precision"))]
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         match self.n {
-            N::PosInt(u) => Display::fmt(&u, formatter),
-            N::NegInt(i) => Display::fmt(&i, formatter),
-            // Preserve `.0` on integral values, which Display hides
-            N::Float(f) => Debug::fmt(&f, formatter),
+            N::PosInt(u) => formatter.write_str(itoa::Buffer::new().format(u)),
+            N::NegInt(i) => formatter.write_str(itoa::Buffer::new().format(i)),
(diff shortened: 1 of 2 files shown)


src/number.rs
```diff
@@ -294,7 +294,8 @@ impl Display for Number {
         match self.n {
             N::PosInt(u) => Display::fmt(&u, formatter),
             N::NegInt(i) => Display::fmt(&i, formatter),
-            N::Float(f) => Display::fmt(&f, formatter),
+            // Preserve `.0` on integral values, which Display hides
+            N::Float(f) => Debug::fmt(&f, formatter),
```
dtolnay, line 298: Actually we should just serialize using `ryu` and `itoa`, like the Serializer does, or maybe even go through the Serializer itself. `std::fmt` produces a worse representation for floats, like `120000000000000000000000000000000000000000` instead of `1.2e41`.

iitalics:
shouldn't this be a more significant version bump than patch? it has broken a unit test of mine. nothing extreme but was not expecting to see this behavior change

Directory:
rustaceans are afraid of MAJOR semvar version changes. it's like kryptonite to them.

alteous:
Could this be configurable? I have a slightly unusual use case where I need to represent ones' complement integer, i.e., signed integers with a negative zero. I'm using `f64::serialize` but that means integers are displayed as `1.0` instead of `1` which looks weird in this context.

iitalics:
@alteous it looks like you could configure some things through the [`Formatter` trait](https://docs.rs/serde_json/latest/serde_json/ser/trait.Formatter.html)

alteous:
@iitalics that looks promising, thanks! I'll give a try.

Issue #202, opened 2017-01-19
Support deserializing f32 and f64 from null


dtolnay:
Currently f32::NAN and f64::NAN get serialized as null but fail to deserialize back as NAN.

cc @sfackler

sfackler:
It's a bit awkward since inifinity and -infinity also go to null :( Yay JSON

Schultzer:
Could we make a trait wrapper? 

It could be something like `f64PosInfinity/f64NegInfinity` or just `f64Infinity `,`f64NaN`?

cdbrkfxrpt:
Hi,

is there a canonical way of dealing with this?

DanielJoyce:
This JUST bit me. Was tracking down a bug, wandering why I got nulls, then looking at the code, wondering why it wasn't blowing up or why I wasn't seeing NaN in json....

pkolaczk:
I've just got bitten by that as well.

patrickelectric:
What is the status ?

DanikVitek:
Same issue

dnut:
Since I have control over the struct used for both serialization and deserialization, I dealt with this by applying `#[serde_as(as = "DisplayFromStr")]` to the float fields.

```rust
use serde::{Deserialize, Serialize};
use serde_with::{serde_as, DisplayFromStr};

#[serde_as]
#[derive(Deserialize, Serialize)]
pub struct Foo {
    #[serde_as(as = "DisplayFromStr")]
    pub my_float: f64,
}


#[test]
fn serde_json_f64_display_from_string() {
    assert!(test_round_trip(f64::NAN, "NaN").is_nan());
    assert_round_trip_eq(f64::NEG_INFINITY, "-inf");
    assert_round_trip_eq(f64::INFINITY, "inf");
    assert_round_trip_eq(1.1, "1.1");
    assert_round_trip_eq(-100.0, "-100");
    assert_round_trip_eq(0.0, "0");
    assert_round_trip_eq(f64::MIN, "-179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");
[...]

PR #921, opened 2022-08-24
Add support for 'nan'

patrickelectric:
This PR allows the usage of 'nan' in json, like `json!(nan)`.
With that.. null could be used as +/-infinity, and that can help people that uses serde serialization for more complex data types and protocol standards. 

To explain a bit better my use case, I'm using serde with json to create a REST API based the [mavlink](https://mavlink.io/en/) protocol standard, but some messages uses `nan` to define some specific behavior for some sensors. 

PS: since this needs `std::f64::NAN` to use it, should I move it as a feature ?

Helps #202

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -263,10 +263,22 @@ impl<'de, R: Read<'de>> Deserializer<R> {
         let err = match self.peek_or_null().unwrap_or(b'\x00') {
             b'n' => {
                 self.eat_char();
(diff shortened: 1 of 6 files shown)


dtolnay (commented):
`nan` is not a valid JSON value, so I don't think it would be appropriate for a JSON parser to accept in the input or produce it. I would prefer not to make this change.

PR #928, opened 2022-09-05
arbitrary_precision without std

kvinwang:
The feature `arbitrary_precision` didn't compile without the feature `std` enabled. Let's fix it.


dtolnay (approved):
Thanks!

PR #930, opened 2022-09-11
get_concurrent values

KM8Oz:
```
/// Similar to `get` method but fast to get deep json Value
    /// ```
    /// # use serde_json::json;
    ///
    /// let object = json!({ "A": { "B": { "C": "foo" } } }); // true
    ///
    /// assert_eq!(*array.get_concurrent("A.B.C").unwrap(), json!("foo")); // true
    /// assert_eq!(*array.get_concurrent("A.B.C.D").unwrap(), json!("foo")); // true also
    /// ```
```
```
    pub fn get_concurrent(&self, list: &str) -> Option<Value> {
        let rest: Value = Value::default();
        for item in list.split(".").into_iter() {
            let rest = match self.0.get(item) {
                Some(m) => Some(m),
                None => Some(&self.0),
            };
        }
        Some(rest)
    }
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
   schedule: [cron: "40 1 * * *"]
 
+permissions:
+  contents: read
+
 env:
   RUSTFLAGS: -Dwarnings
 


kangalio:
I suspect this won't be merged as-is. It is a rather opinionated feature for such a fundamental crate.

Also, it's not ready needed: `object["A"]["B"]["C"]` works.

KM8Oz:
I made some changes on it so it will fit my needs. Thanks overall

dtolnay:
This looks like an ad-hoc redesign of JSON Pointer ([RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901)). I would prefer not to include this in serde_Json.

PR #938, opened 2022-10-15
feat: impl `From<[(K, V); N]>` for `Map` and `Value`

paul-soporan:
This PR implements `From<[(K, V); N]>` for `Map` and `Value`, enabling things like:
```rs
let map = Map::from([("lorem", 40), ("ipsum", 2)]);
```
or
```rs
let v = [("lorem", 40), ("ipsum", 2)];
let x: Value = v.into();
```

Relevant PRs:
- https://github.com/rust-lang/rust/pull/84111 implemented it for the collections in `std`
- https://github.com/bluss/indexmap/pull/205 implemented it for `IndexMap`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -32,6 +32,12 @@ fn main() {
         println!("cargo:rustc-cfg=no_btreemap_remove_entry");
     }
 
+    // Const generics
+    // https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#const-generics-mvp
+    if minor < 51 {
+        println!("cargo:rustc-cfg=no_const_generics");
+    }
+
     // BTreeMap::retain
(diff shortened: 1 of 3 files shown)


martin-g (commented):
The PR does not have any unit tests.

paul-soporan:
I'm new to writing tests in Rust so I apologize if the answer is obvious:

From what I could find, it seems like this crate doesn't have any existing unit tests, but only integration tests and documentation tests.

The example I added to `From<[(K, V); N]> for Value` is already checked as a doc test, and I could add a doc test to the impl for `Map` too. Is there any advantage to writing unit tests instead (especially when all other `From` `impl`s are only checked via doctests)?

PR #939, opened 2022-10-17
Escape `DEL` character as `\u007f`

sashka:


dtolnay:
You should plug in a formatter that escapes exactly what you want escaped. Something like the following. You can delegate the other Formatter methods to serde_json::ser::PrettyFormatter if you also want indentation.

```rust
use serde::Serialize;
use serde_json::json;
use serde_json::ser::CharEscape;
use std::io::{self, Write};

struct EscapedDelFormatter;

impl serde_json::ser::Formatter for EscapedDelFormatter {
    fn write_string_fragment<W>(&mut self, writer: &mut W, fragment: &str) -> io::Result<()>
    where
        W: ?Sized + Write,
    {
        for (i, fragment) in fragment.split('\x7f').enumerate() {
            if i > 0 {
                self.write_char_escape(writer, CharEscape::AsciiControl(0x7f))?;
            }
            writer.write_all(fragment.as_bytes())?;
        }
        Ok(())
    }
}

fn main() {
    let value = json!({ "thing": "as\x7fdf" });
    let mut buf = Vec::new();
    let mut ser = serde_json::Serializer::with_formatter(&mut buf, EscapedDelFormatter);
    value.serialize(&mut ser).unwrap();
    buf.push(b'\n');
    io::stdout().write_all(&buf).unwrap();
}
```

PR #940, opened 2022-10-19
Add direct support for i128 and u128 via itoa

Lucretiel:
`itoa` natively and unconditionally supports `i128` and `u128`, so we can use it instead of `to_string` on those types.

This change should have no publicly visible behavior changes. It was previously impossible to customize behavior for 128 bit ints in `Formatter`, so there aren't any public facing behavior changes introduced by the addition of new methods to `Formatter`.

The MSRV for `serde_json` is 1.36, so there should be no need to conditionally compile anything in this MR. This is affirmed by the `itoa` dependency, which unconditionally supports 128 bit integers.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -121,7 +121,7 @@ where
     serde_if_integer128! {
         fn serialize_i128(self, value: i128) -> Result<()> {
             self.formatter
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #950, opened 2022-11-14
Simplify number parsing.

01mf02:
This PR simplifies number parsing by making functions for a few reoccurring patterns.

I did not benchmark the impact of these changes.


dtolnay (commented):
I think this was all right as previously written. I get the impression that the abstractions introduced here are focused on minimizing line count, not optimizing understandability, and I don't think they end up serving as helpful abstractions.

Thanks anyway for the PR!

PR #952, opened 2022-11-17
Add clone_from and use IndexMap::extend

Lucretiel:
This PR adds `clone_from` to `Map`'s `Clone` implementation, so that clones can take advantage of existing storage. Additionally, `Map::append` now uses `IndexMap::extend`, to take advantage of reservation logic.


dtolnay (approved):
Thanks!

PR #967, opened 2022-12-21
Add support for arbitrary crate to Value.

SolidTux:
This adds the option to derive the Arbitrary trait (defined in the [arbitrary crate](https://github.com/rust-fuzz/arbitrary/)) for `serde_json::Value`, gated behind the `arbitrary_value` feature.


dtolnay (commented):
Thank you for the PR, but I am not interested in adding this.

SolidTux:
Any specific reason for this (not that there needs to be one)? Would there be some other form of implementation that you would be interested in?

As it's a small change I can maintain the fork for my personal use, but this feature seems to be quite common to have (e.g. indexmap and chrono have it).

Issue #199, opened 2017-01-18
Implement Serialize for Errors


radix:
When I'm implementing an API server using JSON for message passing, I want to be able to send the client the details of the error so that client developers can better debug their implementations. It'd be nice if I could send those details in a structured form, instead of just the Debug representation.

dtolnay:
Seems reasonable. To clarify - you mean specifically the `serde_json::Error` type and not `std::error::Error` types in general, right?

radix:
@dtolnay yep, just the serde_json one. I can have a crack at implementing this.  Do you think it will be as simple as deriving the Serializable trait on the various error types?

dtolnay:
That would tend to give you a pretty verbose representation like:

```json
{
    "Syntax": [
        {
            "MissingField": "f"
        },
        10,
        20
    ]
}
```

I think we should instead implement Serialize to serialize just the Display representation:

```json
"missing field 'f' at line 10 column 20"
```

and implement Deserialize to go through `serde::de::Error::custom` to convert the string back to an Error. In #184 I intend to hide the internal representation anyway so it becomes just an implementation detail / performance knob.

radix:
That representation is actually exactly what I wanted :) (although maybe in an object instead of an array, so `10` and `20` are associated with some descriptive keys).

If I just wanted the display, I could just `to_string(format!("{}", e))`.

radix:
But indeed, any structured serialization would have to wait on #184.

dtolnay:
Can you explain the use case where you need the structured representation? As far as I could tell from some GitHub searches, nobody is currently pattern-matching on our error type ("if it is a MissingField, do this, if it is an InvalidType, do this, etc"). I can understand if you need just the position information; would this be okay?

```json
{
  "message": "missing field 'f'",
  "line": 10,
  "column": 20
}
```

The thing I would like to avoid is committing to a particular error representation in our public API if nobody is using it. We have had problems with that in the past in terms of not being able to add error variants for new error situations because that would be a breaking change.

PSeitz:
Some snippet from the failed position would also be nice, that's what you probably need at the end

svenstaro:
@dtolnay Hey, any chance you could pick this up? This would be quite neat to allow for proper error feedback for users when something fails to deserialize when it's sent to one's API. I'm not really sure how to work around this since I can't implement traits for foreign crates so it would ideally be in `serde_json` itself.

dtolnay:
[...]

PR #971, opened 2023-01-11
error: Implement Serialize

ernestask:
This commit implements serialization for the Error struct that simply defers to the Display implementation.

Fixes https://github.com/serde-rs/json/issues/199

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -382,6 +382,15 @@ impl ser::Error for Error {
     }
 }
 
+impl ser::Serialize for Error {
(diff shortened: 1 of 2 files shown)


ernestask:
@dtolnay, seems like you’re the only person to ping about PRs. :pray:

I saw there was disagreement about using a structured representation. My use case is simply embedding serialization errors, so just the `Display` implementation suffices.

dtolnay (commented):
Thanks for the PR!

I would prefer if you would handle this use case using `serialize_with` attributes instead. #199 discusses downsides with the approach of this PR.

PR #987, opened 2023-02-15
docs: update ambiguity

### LINKED ISSUE #986: Unclear documentation
https://github.com/serde-rs/json/blob/0ebeede28a0d5f0f07f18124078f55d62b180a2d/src/ser.rs#L2137

Wondering to myself, "why does `derive(Serialize)` not pre-determine `serde_json::to_string(..)` to be safe? Then I checked the docs. Turns out I just forgot users can define serializers (which will never be guaranteed safe).

But what didn't make sense to me was the word "map" here. What is a "map" in rust? Is this a HashMap? a BTreeMap?

I put together a very small PR to fix the ambiguity.

simbleau:
Closes #986

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -2134,7 +2134,7 @@ where
 /// # Errors
 ///
 /// Serialization can fail if `T`'s implementation of `Serialize` decides to
-/// fail, or if `T` contains a map with non-string keys.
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks for the PR! I think this is all right as written though.

Issue #972, opened 2023-01-12
serde_json::Error does not implement clone


rdelfin:
Is there a particular reason why the `Error` type in this crate does not implement the `Clone` trait? I'm working on a problem where I basically send a `Result` type over a `tokio::sync::broadcast` channel, but unfortunately I can't make use of this since the Error type has a nexted JSON error without encapsulating the error in an Rc/Arc or some other similar container type. Thanks!

dtolnay:
It's not usual for error types in Rust libraries to implement `Clone`.

Please use `Arc<serde_json::Error>` if both sides of the channel need ownership of the same JSON error.

PR #990, opened 2023-03-05
Implement Clone trait for serde_json::Error

bbergeron0:
Related requests: #972 and #335.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,6 +12,7 @@ use std::error;
 
 /// This type represents all possible errors that can occur when serializing or
 /// deserializing JSON data.
+#[derive(Clone)]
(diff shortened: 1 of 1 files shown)


bbergeron0:
A brief investigation regarding the failing checks showed most alternative solution exclusively compatible to either pre-nightly builds or nightly builds. To pass the checks, the PR needs to implement a backward-compatible solution for cloning `std::io::Error`.

dtolnay (commented):
I am not interested in having a lossy `Clone` impl (which loses the error's `raw_os_error()`, `source()`, `downcast`/`get_ref` behavior, etc). A clone impl that behaves like this would be highly unexpected.

PR #994, opened 2023-03-07
Feature: JSONC Deserialization

nathanhammond:
I propose that `serde_json` support _reading_ (but not tokenizing or serializing) JSONC documents when opting in to a feature.

This is a feature request, but I've included a sample working implementation as a straw proposal. My reason for doing it this way is to demonstrate the minimal scope of the change and how it can have zero impact on those _not_ using the feature.

Why not use, say, https://crates.io/crates/jsonc-parser? Guaranteed consistency with the Rust community's reference implementation and its performance and security guarantees is extremely valuable.

Notes:
- The entire integration point to existing code is addition of a single new match arm.
- The feature name is bikesheddable, I just chose something close.
- There's one "advances the reader index too far" problem noted, but if `serde_json` actually wants to consider this feature I'll go back and adjust the implementation to solve it.

Related: #168

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,3 +89,7 @@ raw_value = []
 # overflow the stack after deserialization has completed, including, but not
 # limited to, Display and Debug and Drop impls.
 unbounded_depth = []
+
+# Enable the deserializer to treat comments in JSONC as whitespace. Comments are
+# not tokenized or serialized.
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
Per the issue you linked, I'm not interested in this for this library.

PR #1002, opened 2023-03-20
Pretty print indent

jwpjrdev:
currently, if a user of the library wants to serialise to a pretty JSON string with an indent other than 2 spaces, this is the way to go about it:
```rs
let value = ...;

let mut buf = Vec::new();
let formatter = serde_json::ser::PrettyFormatter::with_indent(b"    ");
let mut ser = serde_json::Serializer::with_formatter(&mut buf, formatter);
value.serialize(&mut ser)?;
let pretty_json = String::from_utf8(buf)?;
```
not fun.

this PR adds a couple methods, culminating in `serde_json::to_string_pretty_indent`. it does the exact same thing as `to_string_pretty` but it allows you to pass an indent such as `b"    "` down the chain to `PrettyFormatter`. this can be used like so:
```rs
let value = ...;

let pretty_json = serde_json::to_string_pretty_indent(&json, b"    ")?;
```

i have not written any tests because i poked around and saw that there were a lot of things going on. some help with that would be greatly appreciated. i already tested the additions in another project of mine to ensure that it does function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -377,10 +377,10 @@ pub use crate::de::{from_slice, from_str, Deserializer, StreamDeserializer};
 #[doc(inline)]
 pub use crate::error::{Error, Result};
 #[doc(inline)]
-pub use crate::ser::{to_string, to_string_pretty, to_vec, to_vec_pretty};
+pub use crate::ser::{to_string, to_string_pretty, to_string_pretty_indent, to_vec, to_vec_pretty, to_vec_pretty_indent};
 #[cfg(feature = "std")]
 #[doc(inline)]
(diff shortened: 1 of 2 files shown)


jwpjrdev:
i'm also curious—why must a `&[u8]` be provided? wouldn't a `usize` be better?

that way, instead of providing `b"    "`, you could just provide `4`. i continued that trend in this PR but i feel like it might be useful to add another method alongside `to_string_pretty_indent` that simply uses `usize` because that's what 99% of people will be using.

dtolnay:
Non-space indents? `b"\t"`

PR #1003, opened 2023-03-26
Enhance operation on untyped json through operator overload based on json pointer

lymslive:
Although it is wonderfull to deserilize json to strong type struct, sometimes it has to deal with untyped json, especially when the input json is too loose or too freely to fit a struct hierarchy. So I devolop a solution for such case.

I summarize the following basic operation on untyped json tree:

1. Locate and refer to some node in the tree by json pointer syntax, using operator `/` overload.
2. Read value from leaf node with default fallback, using operator `|`, image as `get_or` method.
3. Write value to json node as container of common data type, using operator `<<`, image as push item to array or object, or overwrite put value to scalar node.
4. Some advanced modification on json tree as a whole, using operator `|`, pipe to function or closure. 

All the operators above can be chained as return json pointer, except case 2 which only return primitive scalar value.

Example from that added to `README.md`:

```rust
let v: Value = json( {
        "name": "John Doe",
        "age": 43,
        "phones": [
            "+44 1234567",
            "+44 2345678"
        ]
    });

let age_ptr = &v / "age";
let age_val = age_ptr | 0; // got 43, or default 0 if something wrong.
let first_phone = &v / "phones" / 0 | ""; // got &str
assert_eq!(first_phone, "+44 1234567");

// modify some node
let _ = &mut v / "age" << 44;
let _ = &mut v / "phones" << ["+44 3456789"];
let _ = &mut v << ("sex": "male");
```

And there is also fully documantation example and tests.

Because it is impossible to overload operator for `serde_json::Value` outside the crate, but at most can only to overload operator for new struct for json pointer. So I create this pull request to support json operator natively as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,40 @@ unrecognized fields in the input data. The compiler is powerless to help you
 when you make a mistake, for example imagine typoing `v["name"]` as `v["nmae"]`
 in one of the dozens of places it is used in your code.
 
+### JSON pointer for untyped JSON values and operator overload
+
+Json pointer is an usefull tool to refer to and operate on individual json
+node. There is a method `.pointer()` for `serde_json::Value` to create json
+pointer, and beside that, `/` is overloaded as path operator to yield pointer
+more conveniently and visually，and it can be chained to point to furthur
+deeper node along the json tree, then there is other operator to read from or
+write to the node it pointed.
+
+```rust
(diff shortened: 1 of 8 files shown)


dtolnay (commented):
Thank you for the PR, but I would prefer not to support this mechanism in this library.

Issue #340, opened 2017-07-14
How to get the original string value of floating point number?


shelbyd:
I'm building an application that needs precise numbers. An API that I'm getting data from sends their numbers in JSON as raw numbers. I need the value as transmitted over the wire so I can build a num::Ratio from it without rounding errors.

```rust
extern crate serde;
extern crate serde_json;
#[macro_use] extern crate serde_derive;

#[derive(Deserialize, Eq, PartialEq, Debug)]
struct Test {
    foo: String
}

fn main() {
    let input = r#"{"foo": 1.01}"#;
    let t: Test = serde_json::from_str(input).unwrap();
    assert_eq!(t, Test { foo: "1.01".to_owned() });
}
```

The only thing I've thought that would work is to make my own struct that implements Deserializer and delegates to serde_json::Deserializer unless there are numbers. The problem is that I cannot interact with the read or call the parse_* methods of serde_json::Deserializer.

Am I going in the right direction? Am I missing something obvious?

I've seen https://github.com/serde-rs/serde/issues/185, but that doesn't seem to be in 1.0.2.

Next, I'm probably going to fork the repo and update the code to fit my purpose. But I hate doing that, so I'm checking for another option.

dtolnay:
This is not supported. If you are forking the code, there is a working implementation in #252 to get you started.

shelbyd:
For posterity, I got this working with the current release of serde/serde_json.

```rust
extern crate serde;
extern crate serde_json;
#[macro_use] extern crate serde_derive;

use serde::de::{Deserialize, Deserializer};
use serde_json::Number;

#[derive(Deserialize, Eq, PartialEq, Debug)]
struct Test {
    #[serde(deserialize_with = "foo")]
    foo: String
}

fn foo<'de, D: Deserializer<'de>>(d: D) -> Result<String, D::Error> {
    let n: Number = Deserialize::deserialize(d)?;
    Ok(n.to_string())
}

fn main() {
    let input = r#"{"foo": 1.01}"#;
    let t: Test = serde_json::from_str(input).unwrap();
    assert_eq!(t, Test { foo: "1.01".to_owned() });
}
```

My best guess is that the rust compiler notices I'm essentially calling `"1.01".parse::<f64>().to_string()` and removes the `parse` & `to_string` calls. It also works if I change the type from `Number` to `f64` inside the `foo` function.

I don't really like relying on compiler magic, but it's the best option I've found so far.

dtolnay:
[...]

PR #1017, opened 2023-04-27
Add impl From<Number> for String

utkarshgupta137:
I want to parse a JSON such as `{"float": 1.23456}`, but I don't want to lose precision & keep it as a string. Currently, I've enabled the arbitrary_precision feature, but I've to call the `to_string()` function to get the string, which allocates a new string & then uses `Formatter`. Adding this impl allows me to avoid extra heap allocations for getting the string.

Related: https://github.com/serde-rs/json/issues/340


utkarshgupta137:
For context, we have got a JSON with a vec of structs (around 10-20 elements in a message), each of which contains about 10-12 floats with 10-14 digits of precision. By making this change, we're able to reduce the deserialization time of the message by 30%.

utkarshgupta137:
@dtolnay, sorry for pinging, but any hope for this to be accepted?

dtolnay (commented):
I would prefer not to add this because I have a hard time seeing any situation where this API would constitute the best solution. to_string() is almost always going to be fine, and the rest of the time, deserializing to &amp;RawValue would tend to perform better than this PR by not individually heap-allocating every individual number.

utkarshgupta137:
But `RawValue` is only for maps as far as I can tell from `impl Deserialize`. How can I get only the float part as a string?

PR #1021, opened 2023-05-18
Add an example of producing a Null in a json! literal

ndmitchell:
I tried to do this, and while you can piece together the way to do this from the interpolation comment, it's the one piece of direct JSON syntax that isn't immediately obvious.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -2,7 +2,6 @@
 ///
 /// ```
 /// # use serde_json::json;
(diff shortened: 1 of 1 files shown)


ndmitchell:
That good to know - I should have probably guessed that worked, but I didn't. Adjusted the example to use `null` directly.

dtolnay (approved):
Thanks!

PR #1027, opened 2023-06-17
Allow `f32` and `f64` map keys

overdrivenpotato:
Now we allow objects with `f32` or `f64` string-serialized keys. For example:

```json
{ "1.123": "foo", "2.0": "bar" }
```

The `test_deny_float_key` test has been inverted, and the `deserialize_integer_key` macros have been renamed to `deserialize_numeric_key` in order to support `f32` and `f64`.

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -2118,7 +2118,7 @@ struct MapKey<'a, R: 'a> {
     de: &'a mut Deserializer<R>,
 }
 
-macro_rules! deserialize_numeric_key {
+macro_rules! deserialize_integer_key {
     ($method:ident => $visit:ident) => {
         fn $method<V>(self, visitor: V) -> Result<V::Value>
         where
@@ -2132,6 +2132,25 @@ macro_rules! deserialize_numeric_key {
                 (Err(_), Reference::Borrowed(s)) => visitor.visit_borrowed_str(s),
                 (Err(_), Reference::Copied(s)) => visitor.visit_str(s),
             }
+        }
+    };
+}
+
+macro_rules! deserialize_float_key {
+    ($method:ident) => {
+        fn $method<V>(self, visitor: V) -> Result<V::Value>
+        where
+            V: de::Visitor<'de>,
+        {
+            self.de.eat_char();
+            let value = self.de.$method(visitor)?;
+
+            match self.de.peek()? {
+                Some(b'"') => self.de.eat_char(),
+                _ => return Err(self.de.peek_error(ErrorCode::ExpectedDoubleQuote)),
+            }
+
+            Ok(value)
         }
     };
 }
@@ -2155,18 +2174,19 @@ where
         }
     }
 
-    deserialize_numeric_key!(deserialize_i8 => visit_i8);
-    deserialize_numeric_key!(deserialize_i16 => visit_i16);
-    deserialize_numeric_key!(deserialize_i32 => visit_i32);
-    deserialize_numeric_key!(deserialize_i64 => visit_i64);
-    deserialize_numeric_key!(deserialize_i128 => visit_i128);
-    deserialize_numeric_key!(deserialize_u8 => visit_u8);
-    deserialize_numeric_key!(deserialize_u16 => visit_u16);
-    deserialize_numeric_key!(deserialize_u32 => visit_u32);
-    deserialize_numeric_key!(deserialize_u64 => visit_u64);
-    deserialize_numeric_key!(deserialize_u128 => visit_u128);
-    deserialize_numeric_key!(deserialize_f32 => visit_f32);
-    deserialize_numeric_key!(deserialize_f64 => visit_f64);
+    deserialize_integer_key!(deserialize_i8 => visit_i8);
+    deserialize_integer_key!(deserialize_i16 => visit_i16);
+    deserialize_integer_key!(deserialize_i32 => visit_i32);
+    deserialize_integer_key!(deserialize_i64 => visit_i64);
+    deserialize_integer_key!(deserialize_i128 => visit_i128);
+    deserialize_integer_key!(deserialize_u8 => visit_u8);
+    deserialize_integer_key!(deserialize_u16 => visit_u16);
+    deserialize_integer_key!(deserialize_u32 => visit_u32);
+    deserialize_integer_key!(deserialize_u64 => visit_u64);
+    deserialize_integer_key!(deserialize_u128 => visit_u128);
+
+    deserialize_float_key!(deserialize_f32);
+    deserialize_float_key!(deserialize_f64);
 
     #[inline]
     fn deserialize_option<V>(self, visitor: V) -> Result<V::Value>
--- a/src/de.rs
+++ b/src/de.rs
@@ -2118,25 +2118,7 @@ struct MapKey<'a, R: 'a> {
     de: &'a mut Deserializer<R>,
 }
 
-macro_rules! deserialize_integer_key {
-    ($method:ident => $visit:ident) => {
-        fn $method<V>(self, visitor: V) -> Result<V::Value>
-        where
-            V: de::Visitor<'de>,
-        {
-            self.de.eat_char();
-            self.de.scratch.clear();
-            let string = tri!(self.de.read.parse_str(&mut self.de.scratch));
-            match (string.parse(), string) {
-                (Ok(integer), _) => visitor.$visit(integer),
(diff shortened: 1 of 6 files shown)


src/ser.rs
```diff
@@ -1002,12 +1002,38 @@ where
...
+        tri!(self
+            .ser
+            .formatter
+            .begin_string(&mut self.ser.writer)
+            .map_err(Error::io));
+        tri!(self
+            .ser
+            .formatter
+            .write_f32(&mut self.ser.writer, value)
+            .map_err(Error::io));
```
dtolnay, line 1015: This would either produce invalid JSON or nonsense, depending on the formatter, if the float isn't finite.

dtolnay (changes requested):
Would you be able to elaborate on the use case for this?

In general Rust does not make it easy to work with float map keys, because `Eq` is typically required of map keys, and floats do not implement it.

overdrivenpotato:
> Would you be able to elaborate on the use case for this?

This is useful for serializing data types like `BTreeMap<NotNan<f32>, T>`, using a type like [`NotNan`](https://docs.rs/ordered-float/latest/ordered_float/struct.NotNan.html) or [`OrderedFloat`](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html) from the [`ordered-float`](https://crates.io/crates/ordered-float) crate. A type like that could be useful in a few situations, e.g. a sorted map for `timestamp -> event` data, or to maintain an ordered list that allows for insertion *between* integer indices.

src/ser.rs
overdrivenpotato: I've now added checks to ensure those values are finite, and tests to ensure serialization with `NaN`, `inf`, and `-inf` keys returns an error.

src/de.rs
```diff
@@ -2155,16 +2155,18 @@ where
...
+    deserialize_numeric_key!(deserialize_i32 => visit_i32);
+    deserialize_numeric_key!(deserialize_i64 => visit_i64);
+    deserialize_numeric_key!(deserialize_i128 => visit_i128);
+    deserialize_numeric_key!(deserialize_u8 => visit_u8);
+    deserialize_numeric_key!(deserialize_u16 => visit_u16);
+    deserialize_numeric_key!(deserialize_u32 => visit_u32);
+    deserialize_numeric_key!(deserialize_u64 => visit_u64);
+    deserialize_numeric_key!(deserialize_u128 => visit_u128);
+    deserialize_numeric_key!(deserialize_f32 => visit_f32);
+    deserialize_numeric_key!(deserialize_f64 => visit_f64);
```
dtolnay, line 2169: `.parse()` is not appropriate to use here for floats. That accepts a bunch of syntax that is not considered a valid float in json. For example `.0` or `00`. Please make sure this accepts only representations that would be valid f32/f64 outside of a map key.

src/value/ser.rs
```diff
@@ -517,12 +521,20 @@ impl serde::Serializer for MapKeySerializer {
...
+        } else {
+            Err(float_key_must_be_finite())
+        }
     }
 
-    fn serialize_f64(self, _value: f64) -> Result<String> {
-        Err(key_must_be_a_string())
+    fn serialize_f64(self, value: f64) -> Result<String> {
+        if value.is_finite() {
+            Ok(value.to_string())
```
dtolnay, line 534: Please use the same logic used elsewhere in the crate for printing floats. A `Map<NotNan<f64>, NotNan<f64>>` should always produce the same digits in the keys and values. As written here, this uses std::fmt's representation, which is different from what we do in JSON. For example in the case of 1e-20.

dtolnay (changes requested):

src/value/ser.rs
overdrivenpotato: I've now updated this to use `CompactFormatter` with a `Vec<u8>` buffer, to keep formatting methods consistent. Would it make more sense to directly use `ryu` here?

src/de.rs
overdrivenpotato: I've updated this to delegate to the parent `Deserializer` via `self.de.method(...)`. This unfortunately makes the error messages a bit worse, as we now get something like:

```
expected value at line 1 column 3
```

Instead of:

```
"invalid type: string \"x\", expected f32 at line 1 column 4"
```

I noticed a similar issue here is present with integer keys, [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9d8c47583ecfb7e0baf8306fdd082f50) works on `serde_json` `master`:

```
extern crate serde_json;
extern crate serde;

fn main() {
    dbg!(serde_json::from_str::<std::collections::HashMap<i32, i32>>(r#"{"00":0}"#));
}
```

Is it worth updating this as well?

src/value/ser.rs
dtolnay: It would make more sense to directly use ryu.

src/value/de.rs
```diff
@@ -1146,16 +1146,18 @@ impl<'de> serde::Deserializer<'de> for MapKeyDeserializer<'de> {
...
+    deserialize_numeric_key!(deserialize_i32 => visit_i32);
+    deserialize_numeric_key!(deserialize_i64 => visit_i64);
+    deserialize_numeric_key!(deserialize_i128 => visit_i128);
+    deserialize_numeric_key!(deserialize_u8 => visit_u8);
+    deserialize_numeric_key!(deserialize_u16 => visit_u16);
+    deserialize_numeric_key!(deserialize_u32 => visit_u32);
+    deserialize_numeric_key!(deserialize_u64 => visit_u64);
+    deserialize_numeric_key!(deserialize_u128 => visit_u128);
+    deserialize_numeric_key!(deserialize_f32 => visit_f32);
+    deserialize_numeric_key!(deserialize_f64 => visit_f64);
```
dtolnay, line 1160: Same problem with `parse` occurs here.

dtolnay (changes requested):

src/de.rs
dtolnay: Yes, good call. I filed #1034 to follow up.

src/value/de.rs
overdrivenpotato: This has been resolved by using `crate::from_str` instead of `str::parse` for float keys, and now also integer keys.

src/value/ser.rs
overdrivenpotato: Updated to use `ryu`.

src/de.rs
overdrivenpotato: In an effort to patch #1034, I've also updated integer key parsing to use `Deserialize::*`. This still exhibits the above error degredation.

Would it make sense here to parse the key into an intermediate string buffer, and then deserialize the integer value using `crate::from_str`? This would result in the original error messages for unexpected string keys, but may have a performance impact.

dtolnay: This needs a check for leading whitespace, otherwise it would end up accepting potentially invalid JSON syntax.

For example this is not a legal JSON document:

```json
{"
1":null}
```

```rust
// [dependencies]
// ordered-float = { version = "3", features = ["serde"] }
// serde_json = "1"

use ordered_float::NotNan;
use std::collections::BTreeMap;

fn main() {
    let invalid_json = r#"
        {"
        1":null}
    "#;

    let map: BTreeMap<NotNan<f64>, ()> = serde_json::from_str(invalid_json).unwrap();
    println!("{:?}", map);
}
```

dtolnay (changes requested):
The serialization looks good!

I left one more comment up in https://github.com/serde-rs/json/pull/1027#pullrequestreview-1520936871 about the deserialization.

src/de.rs
dtolnay: That'll fix the error message problem too. JSON numbers must start with `0-9` or `-`. If the upcoming character is not one of those, you can avoid calling the function that would produced the bad message in this situation, and directly create a useful "invalid value" error instead.

overdrivenpotato: I've patched this with a check for whitespace characters. I don't quite follow what you mean about fixing the error, the misleading text is still present, no? E.g. this example:

```rust
fn main() {
    dbg!(serde_json::from_str::<std::collections::HashMap<i32, i32>>(r#"{"abc":0}"#));
}
```

Previously would print `invalid type: string "abc", expected i32 at line 1 column 6`, but after this changeset it prints `expected value at line 1 column 3`.

dtolnay (approved):
Thank you!

src/de.rs
dtolnay: This is what I was thinking of doing to give a more meaningful error: #1035.

PR #1045, opened 2023-07-26
Implement IntoDeserializer for &Value

### LINKED ISSUE #423: Implement IntoDeserializer for Value and &Value
An `IntoDeserializer` impl would provide a way to deserialize from these types while constructing exactly the right error type on failure, rather than constructing a `serde_json::Error` and mapping it to a custom error of the target type through its `Display` impl.

```rust
fn f<'de, T, D>(deserializer: D) -> Result<T, D::Error>
where
    T: Deserialize<'de>,
    D: Deserializer<'de>,
{
    let mut v = Value::deserialize(deserializer)?;

    /* some manipulation of v */

    T::deserialize(v).map_err(serde::de::Error::custom)
}
```

```diff
- T::deserialize(v).map_err(serde::de::Error::custom)
+ T::deserialize(v.into_deserializer())
```

ZetaNumbers:
Closes #423

Allows to deserialize from something like `&HashMap<String, Value>`

### DIFF
--- a/src/value/de.rs
+++ b/src/value/de.rs
@@ -482,6 +482,14 @@ impl<'de> IntoDeserializer<'de, Error> for Value {
     }
 }
 
+impl<'de> IntoDeserializer<'de, Error> for &'de Value {
+    type Deserializer = Self;
+
+    fn into_deserializer(self) -> Self::Deserializer {
+        self
+    }
+}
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1062, opened 2023-09-03
Simplify/remove build.rs following the bump to 2021 edition

osiewicz:
Since serde-json's MSRV is 1.56 (due to using a 2021 edition), most of the hidden features enabled in build.rs shouldn't ever get enabled during the build. They mostly indicated whether some of the BTreeMap methods were available.

The last one for limb_width can be implemented with a #[cfg] or two. it is a separate commit in case this change is not welcome - I see how it complicates code slightly.


src/map.rs
```diff
@@ -153,44 +152,7 @@ impl Map<String, Value> {
         String: Borrow<Q>,
         Q: ?Sized + Ord + Eq + Hash,
     {
-        #[cfg(any(feature = "preserve_order", not(no_btreemap_remove_entry)))]
         return self.map.remove_entry(key);
```
dtolnay, line 155:
```suggestion
        self.map.remove_entry(key)
```

dtolnay (changes requested):
Thanks! The `retain` and `get_key_value` and `remove_entry` changes look good.

I would prefer not to take the LimbConfig change. I think that code will be all right as is.

dtolnay (approved):
Thanks!

Issue #1064, opened 2023-09-06
Cannot Deserialize JSON array with empty items to tuple struct


jans2002:
The following is test code;
```rust
use serde::{Serialize,Deserialize};
#[derive(Debug, Deserialize)]
struct Reading(Option<i32>, Option<i32>, Option<i32>, i32,Option<f32>, i32, f32,f64,Option<i32>);

fn main() {
    let data  = "[,,,30,,40,5.5,89.9,]";
    let res: Reading = serde_json::from_str(data).unwrap();
    println!("{:?}", res);
}
```
It will show 
```
Error("expected value", line: 1, column: 2)
```
But it should show :
```
Reading(None, None, None, 30, None, 40, 5.5, 89.9, None)
```

PR #1065, opened 2023-09-06
Update de.rs

jans2002:
Support de-serialize  Option element in JSON data like [,,value] # # #1064


dtolnay (changes requested):
Thanks for the PR!

`[,,value]` is not valid JSON. I would not expect serde_json to successfully deserialize it.

jans2002:
Yes, you're right.  I am mislead because it can be parsed when it is manually inputed in  NodeJS. I just want to cancel the PR.

PR #1091, opened 2023-12-18
Support streaming large JSON arrays and objects

### LINKED ISSUE #404: Provide a way for StreamDeserializer to stream an array of values

luxalpa:
This is a follow-up to #526 . It implements `ArrayDeserializer` and `ObjectDeserializer` APIs and allows for arbitrary nesting. The internals are a bit complex due to the way error handling works. Also, in case an actual error happens, maybe one could think about just making every successive call to the deserializer also return errors (i.e. move it into a Failed state)? I am not sure about that part of the design yet. It is also somewhat adjacent to #70  

It uses lifetimes to ensure only one nested deserializer can be used at a time.

Fixes #404


luxalpa:
I'm new to this whole Open Source thing and this is my first pull request so please let me know if there's anything I need to do (or did wrong) @dtolnay

ivankras:
Is anyone looking into this?

PR #1101, opened 2024-01-05
Add support to `preserve_order` in `no_std`

yukibtc:
This PR add support to preserve order in `no_std` env.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,8 @@ repository = "https://github.com/serde-rs/json"
 rust-version = "1.56"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


dtolnay (changes requested):
Thanks for the PR. I am interested in supporting this, but not with an `ahash` dependency. That crate is problematic (https://github.com/tkaitchuck/aHash/issues/238, https://github.com/tkaitchuck/aHash/pull/239, https://github.com/tkaitchuck/aHash/pull/242) and not well maintained. Would you be able to look into a different implementation?

PR #1102, opened 2024-01-06
[Deserialize] Split out peek logic from seed generic

GnomedDev:
Splits duplicated and somewhat complex logic depending on two generics (`Read` and `seed`) into only depending on one generic (`Read`) cutting debug mode llvm-lines output on my program by:
- `SeqAccess::next_element_seed`: 34232 -> 21752
- `MapAccess::next_key_seed`: 11225 -> 6722

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1192,8 +1192,8 @@ impl<'de, R: Read<'de>> Deserializer<R> {
                     Some(_) => {
                         if accept_comma {
                             return Err(self.peek_error(match frame {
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
Thanks for the PR.

I think extracting a `peek_next_elem` function that is shared between SeqAccess and MapAccess is going to be counterproductive because it interferes with making more of the logic independent of the seed type. This PR is still leaving a significant amount of logic that is duplicated for each seed, as opposed to once per Read type.

Check out #1205 for a different approach that produces 25% smaller `next_element_seed` and `next_key_seed` functions compared to this one.

PR #1105, opened 2024-01-26
fix: Correct spelling error

keienWang:
fix: Correct the only spelling error

After scanning the entire repository, I discovered the only spelling error and fixed it.

- Changed "exponet" to "exponent"


dtolnay (approved):
Thanks!

PR #1113, opened 2024-02-27
Add feature to error on duplicate key in Map<String, Value>

jfehr67:
Add optional feature to error on deserialize of duplicate key in Map<String, Value>. VERY useful if you typically have very large json files that are contributed to by multiple engineers.


DarkKirb:
this is also useful for cryptographic libraries that need to reject duplicate keys in json to avoid canonicalization issues

PR #1119, opened 2024-03-25
Fix missing backtick in doc comments letting markdown think &[u8] is a link to u8

titaniumtraveler:
While reading the documentation/the source code (for the purpose of creating my own `Deserializer` for a different format) I found that in [`serde_json::de::Read`](https://docs.rs/serde_json/latest/serde_json/de/trait.Read.html) `&[u8]` was rendered incorrectly.

So I did a quick `grep` and fixed all I could find by adding backticks.

I also switching the `&[u8]` in the `README.md`/`src/lib.rs` from escapes to backticks for consistency sake. \
(I assume the `README.md` was generated from the `src/lib.rs` with some tool, but if both are changed the same way, it should still be consistent with the output.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ enum Value {
 
 A string of JSON data can be parsed into a `serde_json::Value` by the
 [`serde_json::from_str`][from_str] function. There is also
-[`from_slice`][from_slice] for parsing from a byte slice &\[u8\] and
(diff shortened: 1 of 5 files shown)


dtolnay (approved):
Thank you! Good catch.

Issue #1040, opened 2023-07-14
Supporting no-std and no-alloc


rjsberry:
Would you be interested in reviewing a PR that adds support for no-std + no-alloc? Currently either the std or alloc feature is necessary.

#606 introduced support for no-std + alloc. However, since the alloc feature is still required, it doesn't address users who cannot or prefer not to use an allocator. Allocators can introduce unpredictability in terms of execution times and runtime behavior, which is particularly undesirable for resource-constrained devices. In some cases development guidelines may prohibit allocation altogether. Arguably not be the best choice for an embedded device, JSON does appear in standardized protocols that may be mandated for qualified products.

Users are directed to use [`serde-json-core`](https://docs.rs/serde-json-core) for JSON (de)serialization without the standard library via https://serde.rs/no-std.html. The `serde` and `serde_json` crates are widely adopted and trusted. I think it would greatly benefit the embedded ecosystem if users without access to the standard library and without an allocator could leverage the high-quality implementation of the official JSON crate from the Serde project. It's much easier to make a case for introducing a v1.x.x library with 150M downloads!

I think a few changes would be necessary for the specific no-std + no-alloc build:

1. Configurable recursion (first requested in #334).
2. Simplify `Error`, perhaps just storing category/line/column.
3. Fixed capacity deserializer scratch that is also configurable to users as it will affect stack utilisation.

I believe this can be implemented with little maintenance overhead or any major API differences. For example, implementing 1 and 3 above via environment variables at compile time (perhaps also with an `unstable` feature) keeps the crate API identical across all build variants. Of course the `Debug` and `Display` impls of `Error` would change for no-std + no-alloc, but generally speaking human readable errors would not be a priority for firmware.

As mentioned if you'd be open to a PR I'd definitely like to implement this.

PR #1122, opened 2024-04-15
Make all serialization functions available for no-std

devrandom:
We ran into OOMs in a memory constrained no-std environment, when serializing.  We noticed that `to_writer` was not available for no-std, even though there is nothing preventing it from working.

This PR exposes the relevant serialization functions.  It also exposes the crate's no-std io shim, since types from there appears in the `ser` API.

related to #1040

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ indexmap = { version = "2.2.1", optional = true }
 itoa = "1.0"
 ryu = "1.0"
 serde = { version = "1.0.194", default-features = false }
+core2 = { version = "0.4", optional = true }
 
 [dev-dependencies]
 automod = "1.0.11"
@@ -39,6 +40,9 @@ rustdoc-args = ["--cfg", "docsrs", "--generate-link-to-definition"]
 [package.metadata.playground]
 features = ["raw_value"]
 
+[[example]]
+name = "test"
+required-features = ["use-core2"]
 
 ### FEATURES #################################################################
 
@@ -86,3 +90,5 @@ raw_value = []
 # overflow the stack after deserialization has completed, including, but not
 # limited to, Display and Debug and Drop impls.
 unbounded_depth = []
+
+use-core2 = ["core2"]
--- a/examples/test.rs
+++ b/examples/test.rs
@@ -2,26 +2,29 @@
 //! Run via:
 //!
 //! ```sh
-//! cargo run --example test
-//! cargo run --example test --features alloc --no-default-features
+//! cargo run --example test --features use-core2
+//! cargo run --example test --features alloc,use-core2 --no-default-features
 //! ```
 
 
-use serde_json::io::{Write, Result, Error, ErrorKind};
+use serde_json::alloc_io::{Write, Result, Error, ErrorKind};
 
-struct X {
-}
 
-impl Write for X {
-	fn write(&mut self, _buf: &[u8]) -> Result<usize> {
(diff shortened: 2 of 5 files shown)


dtolnay (changes requested):
Thanks for the PR!

This isn't appropriate to expose in this way, because it makes the "std" feature non-additive. Cargo features are required to be additive. In this PR, one crate could build fine against no-std serde_json but as soon as any other crate in someone's dependency graph enables "serde_json/std", it would cause compilation failures in the first crate, which is not supposed to happen with additive features.

Could you see if it's possible for the "std" feature to remain additive? If not, I would prefer for the no-std use case to be served by a different library. This one does not need to be the only JSON library for serde.

devrandom:
If I understand correctly, you are concerned that if someone refers to the `no_std` version of structs / fns in `serde_json::io`, then they could break if the implementation changes to `std` via feature unification.

While you are correct in general, I don't think that's possible in this case, because the `alloc` version of `serde_json::io` is a strict subset of `std::io`, so something that compiled with the `alloc` version cannot fail to compile once the implementation changes to `std`.  There are no extra fns in the former that are not in the latter, etc. .  I tried to write a failing example, but could not come up with one.

dtolnay:
Some other examples of things that compile only if std is off, and fail non-additively as soon as std is enabled anywhere:

```rust
fn main() {
    let msg = Box::new("...");
    let _ = serde_json::io::Error::new(serde_json::io::ErrorKind::Other, &msg);
}
```

```rust
fn main() {
    match serde_json::io::ErrorKind::Other {
        serde_json::io::ErrorKind::Other => {}
    };
}
```

```rust
fn main() {
    let error = serde_json::io::Error::new(serde_json::io::ErrorKind::Other, "...");
    let _ = std::panic::catch_unwind(|| error);
}
```

```rust
use std::fmt::{self, Debug, Display};

enum MyError {
    JsonIo(serde_json::io::Error),
}

impl Display for MyError {
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        match self {
            MyError::JsonIo(error) => error.fmt(formatter),
        }
    }
}
```

```rust
trait WriteExt: serde_json::io::Write {/* ... */}

impl serde_json::io::Write for &mut dyn WriteExt {
    fn write(&mut self, bytes: &[u8]) -> serde_json::io::Result<usize> {/* ... */}
    fn flush(&mut self) -> serde_json::io::Result<()> {/* ... */}
}
```

Not sure what can be done about this one:

```rust
struct Buffer {/* ... */}

impl serde_json::io::Write for Buffer {
    fn write(&mut self, bytes: &[u8]) -> serde_json::io::Result<usize> {/* ... */}
    fn flush(&mut self) -> serde_json::io::Result<()> {/* ... */}
}

impl core2::io::Write for Buffer {
    fn write(&mut self, bytes: &[u8]) -> core2::io::Result<usize> {/* ... */}
    fn flush(&mut self) -> core2::io::Result<()> {/* ... */}
}
```

devrandom:
OK, good points.

I pushed a commit that changes the approach - the no-std io module is now always exported.

However, this is still non-additive, because `Serializer` uses different `Write` depending on the feature.

I'm not sure there's a complete solution without duplicating `Serializer` for the two cases.

dtolnay (commented):
In that case, I'd prefer for the no-std use case to be served by a different library.

Thanks anyway for the PR!

PR #1127, opened 2024-04-25
impl `Hash` for `Value`

edwardycl:
`IndexMap` is the only type that is used within `Value` that does not implement `Hash`, and it is by design: https://github.com/indexmap-rs/indexmap/issues/288#issuecomment-1804735361

If "preserve_order" feature is not set, `Value::Object` uses `BTreeMap` instead of `IndexMap` so it is possible to auto-impl `Hash` for `Value`.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -8,6 +8,8 @@
 
 use crate::value::Value;
 use alloc::string::String;
+#[cfg(feature = "preserve_order")]
+use alloc::vec::Vec;
 use core::borrow::Borrow;
 use core::fmt::{self, Debug};
 use core::hash::Hash;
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
Cargo features are required to be additive, so enabling a feature cannot result in an impl being removed.

edwardycl:
Now that I think about it, since the type of key is fixed to `String`, `IndexMap` can also be hashed by sorting its key-value in a consistent order. It doesn't break the Eq-Hash relationship, and we can implement hashing independently of features. What do you think?

dtolnay (approved):
Thanks!

For the `cfg(feature = "preserve_order")` case, maybe a better performing implementation would be to hash each entry individually and xor them together. This would make the hash insensitive to order.

```rust
use std::hash::{BuildHasher, RandomState};
use std::sync::OnceLock;

static RANDOM_STATE: OnceLock<RandomState> = OnceLock::new();
let random_state = RANDOM_STATE.get_or_init(RandomState::new);
self.map
    .iter()
    .map(|entry| random_state.hash_one(entry))
    .fold(0, |xor, next| xor ^ next)
    .hash(state);
```

I would be interested to see a benchmark comparing this approach.

PR #1128, opened 2024-05-01
Change serde_json::from_value to use pass-by-reference

rooney:
_If the function needs ownership, you should pass by value. If the function only needs a reference, you should pass by reference._

-- https://stackoverflow.com/a/64814149/342293

`from_value` doesn't need to own its parameter, so it should use pass-by-reference.

### DIFF
--- a/src/value/mod.rs
+++ b/src/value/mod.rs
@@ -987,7 +987,7 @@ where
 ///         "location": "Menlo Park, CA"
 ///     });
 ///
-///     let u: User = serde_json::from_value(&j).unwrap();
+///     let u: User = serde_json::from_value(j).unwrap();
 ///     println!("{:#?}", u);
 /// }
 /// ```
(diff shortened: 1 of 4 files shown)


GnomedDev:
Is this not an obvious breaking change that would require a serde_json 2.0?

rooney:
> Is this not an obvious breaking change that would require a serde_json 2.0?

Hmm yes, any other parts that should also been updated? (Apologies if these should be obvious, still learning)

rooney:
@GnomedDev Redone the fix so now `from_value` accepts both `&Value` and `Value` (maybe the latter can be deprecated?). New test for the `&Value` is on doctest (should I add more?)

dtolnay (commented):
Thanks for the PR, but I prefer the existing signature. The entire point of `from_value` is to enforce a DeserializeOwned bound around `T::deserialize`. If you are going to make the caller write `from_value::<T, &Value>(&v)`, that is now longer than `T::deserialize(&v)` and also enforces a bound that is not correct for deserializing from `&'de Value`, where Deserialize\<'de\> is the correct bound and DeserializeOwned would be too strict.

PR #1131, opened 2024-05-09
Add option to disable array line breaks in `PrettyFormatter`

DervexDev:
This PR makes array line breaks optional using new constructor `PrettyFormatter::with_array_breaks`.

Before:
```json
{
  "size": [
    1,
    2,
    3
  ]
}
```

After:
```json
{
  "size": [1, 2, 3]
}
```


dtolnay (commented):
Thanks for the PR!

I would prefer not to support this in this crate, but a more capable Formatter implementation can be maintained as a separate crate.

PR #1135, opened 2024-05-14
Implement `Deserializer` for `Map<String, Value>` and `&Map<String, Value>`

### LINKED ISSUE #1134: Implement `Deserializer` for `Map<String, Value>` and `&Map<String, Value>`
Sometimes I want to work with a `Value` that I know to be a json object. `Map<String, Value>` is the correct type to use for this but it is not possible to deserialize from a `Map<String, Value>` to a more concrete without going through `Value`. It would be nice if `Map<String, Value>` implemented `Deserializer`  so this all just works.

My actual use case is that I have a `&Map<String, Value>` that I would like to deserialize to a concrete struct, which is harder to come up with a workaround for. With an owned value I could just do `T::deserialize(Value::Object(map))` but that doesn't work for a reference.

swlynch99:
This PR adds `Deserializer` impls for `Map<String, Value>` and `&Map<String, Value>` and `IntoDeserializer` impls for the same. The first commit moves the relevant code from the value `Deserializer` impl into the new map impls. The last two are some small cleanups to use the map deserializer impl directly.

I put the `Deserializer` trait impls in `src/value/de.rs` so that they can reuse the existing helpers there. If you would prefer that they go elsewhere I am happy to move them.

Closes #1134

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -535,6 +535,22 @@ macro_rules! delegate_iterator {
     }
 }
 
+impl<'de> de::IntoDeserializer<'de, crate::Error> for Map<String, Value> {
+    type Deserializer = Self;
+
+    fn into_deserializer(self) -> Self::Deserializer {
+        self
+    }
+}
+
+impl<'de> de::IntoDeserializer<'de, crate::Error> for &'de Map<String, Value> {
+    type Deserializer = Self;
+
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1141, opened 2024-06-17
Add i128 deserialization

druide:
Implemented serde::de::Visitor::visit_i128.
This fixes compatability with Ipld::Integer when using DagJsonCodec.


dtolnay (approved):
Thanks!

PR #1148, opened 2024-06-30
Recover raw values after losing the type

IvanUkhov:
Hello,

I have a use case where a struct with a raw value gets converted into a `Value`, which then needs to be serialized to JSON. Upon conversion, `RawValue` gets downcast to a `Map`, which loses the special serialization of `RawValue`, resulting in a JSON containing an object with one entry whose key equals `raw::TOKEN`.

I am wondering if you would be willing to have code for recovering the “rawness.”

### DIFF
--- a/src/value/ser.rs
+++ b/src/value/ser.rs
@@ -23,6 +23,14 @@ impl Serialize for Value {
             #[cfg(any(feature = "std", feature = "alloc"))]
             Value::Object(m) => {
                 use serde::ser::SerializeMap;
+                #[cfg(feature = "raw_value")]
+                if m.len() == 1 {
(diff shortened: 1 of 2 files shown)


dtolnay (commented):
Thanks for the PR.

Could you give more detail about what your use case is doing to end up with a `Value` like this? `serde_json::to_value` should not be producing such a value, for example.

It would be better if the unit test could be more representative of that use case.

IvanUkhov:
It was a good question that made me think a little bit harder. After going back and forth, I have decided to propose a similar change to `async-graphql`, which was the use case I mentioned:

https://github.com/async-graphql/async-graphql/pull/1554

PR #1149, opened 2024-06-30
Add Map::shift_insert()

joshka:
This method inserts a key-value pair in the map at the given index. If
the map did not have this key present, `None` is returned. If the map
did have this key present, the key is moved to the new position, the
value is updated, and the old value is returned.

This is useful when you want to insert a key-value pair at a specific
position in the map, and is a necessary method when writing a JSON
editor that can mutate the keys in a JSON object.

This is only enabled when the preserve-order feature flag is enabled.

```rust
map.shift_insert(0, "foo", "bar")
```

Mirrors the logic in [`IndexMap::shift_insert()`](https://docs.rs/indexmap/2.2.6/indexmap/map/struct.IndexMap.html#method.shift_insert)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -127,6 +127,17 @@ impl Map<String, Value> {
         self.map.insert(k, v)
     }
 
+    /// Insert a key-value pair in the map at the given index.
+    ///
+    /// If the map did not have this key present, `None` is returned.
+    ///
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1152, opened 2024-07-01
fix: indexmap minimal version with Map::shift_insert()

cforycki:
PR #1149 introduced `Map::shift_insert()` but version of `indexmap` stayed on `2.2.1` which doesn't have this, causing issue on build where `serde_json` is used as a dep.

Updating dep to minimal version of `2.2.3`

(Let me know if it should be `2.2.6` because it's the version in the lock)


dtolnay (approved):
Thanks!

PR #1153, opened 2024-07-02
Correct documentation URL for Value's Index impl.

dpathakj:
The current id doesn't exist, so the link just goes to the top of the docs for Value, rather than to the intended section covering `impl<I> Index<I> for Value`.


dtolnay (approved):
Thanks!

PR #1159, opened 2024-07-26
Move call to tri! out of check_recursion!

purplesyringa:
As far as I can see, all uses of `check_recursion!` except this one don't early-return. The current code decreases `remaining_depth` but doesn't increase it back on visit failure. I don't have a test, but this is probably a bug.


dtolnay (approved):
Thanks!

PR #1171, opened 2024-08-09
Added to_string_mut and to_string_mut_pretty

JohnScience:
```
/// Serialize the given data structure as JSON into the pre-existing [`String``].
///
/// # Errors
///
/// Serialization can fail if `T`'s implementation of `Serialize` decides to
/// fail, or if `T` contains a map with non-string keys.
#[inline]
#[cfg(feature = "std")]
#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
pub fn to_string_mut<T>(s: &mut String, value: &T) -> Result<()>
where
    T: ?Sized + Serialize,
{
    // to_writer() guarantees that it feeds only valid UTF-8 to the writer.
    let buf = unsafe { s.as_mut_vec() };
    let mut writer = std::io::Cursor::new(buf);
    to_writer(&mut writer, value)
}

/// Serialize the given data structure as pretty-printed JSON into the pre-existing [`String`].
///
/// # Errors
///
/// Serialization can fail if `T`'s implementation of `Serialize` decides to
/// fail, or if `T` contains a map with non-string keys.
#[inline]
#[cfg(feature = "std")]
#[cfg_attr(docsrs, doc(cfg(feature = "std")))]
pub fn to_string_mut_pretty<T>(s: &mut String, value: &T) -> Result<()>
where
    T: ?Sized + Serialize,
{
    // to_writer_pretty() guarantees that it feeds only valid UTF-8 to the writer.
    let buf = unsafe { s.as_mut_vec() };
    let mut writer = std::io::Cursor::new(buf);
    to_writer_pretty(&mut writer, value)
}

```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,13 +389,13 @@ pub use crate::de::{from_slice, from_str, Deserializer, StreamDeserializer};
 #[doc(inline)]
 pub use crate::error::{Error, Result};
 #[doc(inline)]
-pub use crate::ser::{to_string, to_string_pretty, to_vec, to_vec_pretty};
+pub use crate::ser::{
+    to_string, to_string_pretty, to_vec, to_vec_pretty, to_writer, to_writer_pretty,
+};
 #[cfg(feature = "std")]
 #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
 #[doc(inline)]
-pub use crate::ser::{
-    to_string_mut, to_string_mut_pretty, to_writer, to_writer_pretty, Serializer,
-};
+pub use crate::ser::{to_string_mut, to_string_mut_pretty, Serializer};
 #[doc(inline)]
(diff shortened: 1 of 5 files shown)


JohnScience:
Ready for the review!

src/lib.rs
```diff
@@ -389,11 +389,13 @@ pub use crate::de::{from_slice, from_str, Deserializer, StreamDeserializer};
 #[doc(inline)]
 pub use crate::error::{Error, Result};
 #[doc(inline)]
-pub use crate::ser::{to_string, to_string_pretty, to_vec, to_vec_pretty};
+pub use crate::ser::{
+    to_string, to_string_pretty, to_vec, to_vec_pretty, to_writer, to_writer_pretty,
```
dtolnay, line 393: https://github.com/serde-rs/json/pull/1122 describes some considerations about exposing no-std to_writer. What's done in this PR is not going to work.

PR #1172, opened 2024-08-11
Parse \uXXXX escapes faster

purplesyringa:
When ignoring *War and Peace* (in Russian), this increases performance from 640 MB/s to 1080 MB/s (+70%).

When parsing into String, the savings are moderate but still significant: 275 MB/s to 320 MB/s (+15%).

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -361,16 +361,14 @@ where
     }
 
     fn decode_hex_escape(&mut self) -> Result<u16> {
-        let mut n = 0;
-        for _ in 0..4 {
-            match decode_hex_val(tri!(next_or_eof(self))) {
-                None => return error(self, ErrorCode::InvalidEscape),
-                Some(val) => {
-                    n = (n << 4) + val;
-                }
-            }
+        let a = tri!(next_or_eof(self));
+        let b = tri!(next_or_eof(self));
+        let c = tri!(next_or_eof(self));
+        let d = tri!(next_or_eof(self));
+        match decode_four_hex_digits(a, b, c, d) {
+            Some(val) => Ok(val),
+            None => error(self, ErrorCode::InvalidEscape),
         }
-        Ok(n)
     }
 
     #[cfg(feature = "raw_value")]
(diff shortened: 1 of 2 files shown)


purplesyringa:
Is changing error precision okay? I couldn't sidestep that without sacrificing performance.

dtolnay:
> Is changing error precision (see the changed test) okay? I couldn't sidestep that without sacrificing performance.

The code previously pointed to a specific one of the expected hex digits:

```haskell
"\u0000\u00#0\u0000"
           ^
```

Preserving that is not important to me.

The new code always points to the 6th byte after the backslash.

```haskell
"\u0000\u00#0\u0000"
             ^
```

Would it be costly to point to the current escape's backslash instead? This would make more sense and hopefully be as simple as offsetting the index by -6 in the error codepath.

```haskell
"\u0000\u00#0\u0000"
       ^
```

purplesyringa:
I can do that easily in the `str` case but not the I/O case. Is that fine?

purplesyringa:
Hmm. Now that I think about it, I'm not sure if I understand how error location works. `position()` is typically called after a call to `next()`, so the error points at the byte *after* the wrong one.

So e.g. here serde-json reports the error immediately after consuming malformed UTF-8:
https://github.com/serde-rs/json/blob/cf771a0471dd797b6fead77e767f2f7943740c98/tests/test.rs#L1080-L1083

And here we say "line 2" after consuming the erroneous `\n`:
https://github.com/serde-rs/json/blob/cf771a0471dd797b6fead77e767f2f7943740c98/tests/test.rs#L1108-L1111

I think throwing up pointing after an invalid escape (i.e. the current behavior) would be more consistent.

dtolnay:
> I'm not sure if I understand how error location works. `position()` is typically called after a call to `next()`, so the error points at the byte *after* the wrong one.

Ideally the position returned by `position()` is supposed to point at the byte most recently produced by `next()`. Separately there is `peek_position()` to return the position of the next byte not yet returned by `next()`.

Sometimes there are codepaths where the byte which caused an error might have come from either `next()` or `peek()`, so the error position won't always be right in that case. Sometimes it's possible to move the error position computation earlier, so that it knows whether `position()` or `peek_position()` is the one to use. It is not intended in general that the reported error positions always refer to the next byte after the bad one.

https://github.com/serde-rs/json/blob/b4bc6436acdf874729a7b425b3c2e48f8b16b398/src/read.rs#L39-L57

dtolnay:
> I can do that easily in the `str` case but not the I/O case. Is that fine?

That would be fine. Something approximate would also be fine, like `column.saturating_sub(6)` in the I/O case, which will be correct almost all of the time with the exception of something grossly malformed like:

```
"\u00
0"
```

dtolnay (approved):
This looks good!

I am interested in error position improvements (not just for `\u` but the other ones you called out too) but that can happen separately if you are interested in looking into it.

purplesyringa:
Thanks!

PR #1173, opened 2024-08-11
Oops, fix skip_to_escape on BE architectures

purplesyringa:
I have belatedly realized that using `leading_zeros` does not, in fact, work on big-endian machines. I have updated the test to trigger the failure (reproduce with `cargo miri test --test test --target powerpc64-unknown-linux-gnu test_control_character_search`) and switched to the slower-but-correct `from_le_bytes` method instead.


dtolnay (approved):
Thanks!

We could add CI coverage for some other architectures, using Miri. There is an example here:

https://github.com/dtolnay/semver/blob/1.0.23/.github/workflows/ci.yml#L103-L110

PR #1174, opened 2024-08-11
Test on BE and 32-bit platforms on CI via Miri

purplesyringa:
As per https://github.com/serde-rs/json/pull/1173#pullrequestreview-2231890262.


dtolnay (approved):
Thank you for following up on that suggestion.

PR #1175, opened 2024-08-12
Speed up \uXXXX parsing and improve WTF-8 handling

purplesyringa:
Altogether, this speeds up \u-encoded War and Peace parsing by 20%. Performance on json-benchmark is slightly affected: there are some 5% improvements and a -1% regression, but I'm willing to write that off as noise from an imperfect benchmark setup.

This PR should be more readable per-commit w/o whitespace changes. In addition to the above, it includes a variation on #877, since it's easier to implement with this design.

### DIFF
--- a/src/read.rs
+++ b/src/read.rs
@@ -898,20 +898,12 @@ fn parse_unicode_escape<'de, R: Read<'de>>(
     validate: bool,
     scratch: &mut Vec<u8>,
 ) -> Result<()> {
-    fn encode_surrogate(scratch: &mut Vec<u8>, n: u16) {
-        scratch.extend_from_slice(&[
-            (n >> 12 & 0b0000_1111) as u8 | 0b1110_0000,
-            (n >> 6 & 0b0011_1111) as u8 | 0b1000_0000,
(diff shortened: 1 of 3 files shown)


src/read.rs
```diff
@@ -882,88 +881,133 @@ fn parse_escape<'de, R: Read<'de>>(
...
+) -> Result<()> {
+    let mut n = tri!(read.decode_hex_escape());
+
+    // Non-BMP characters are encoded as a sequence of two hex
+    // escapes, representing UTF-16 surrogates. If deserializing a
+    // utf-8 string the surrogates are required to be paired,
+    // whereas deserializing a byte string accepts lone surrogates.
+    if validate && n >= 0xDC00 && n <= 0xDFFF {
+        // XXX: This is actually a trailing surrogate.
+        return error(read, ErrorCode::LoneLeadingSurrogateInHexEscape);
```
purplesyringa, line 909: Should I do anything about this?

dtolnay: I would accept a followup PR to change the ErrorCode enum and fix the error message.

dtolnay (approved):
Thank you!

PR #1178, opened 2024-08-19
Optimize Unicode decoding by 1% 🤡

purplesyringa:
This removes two instructions (tops) and frees up one register. This is really, really stupid, and probably does not affect performance in practice (It does on the memory-bound War and Peace benchmark though. By 1%.), but I couldn't stop myself from patching this when I read the code with a fresh mind. Sorry.


dtolnay (approved):
Thanks!

PR #1179, opened 2024-08-19
Add more removal methods to `OccupiedEntry`

GREsau:
`serde_json::Map` has several methods for removing an entry:

- `.remove(key)`
- `.swap_remove(key)` (with preserve_order)
- `.shift_remove(key)` (with preserve_order)
- `.remove_entry(key)`
- `.swap_remove_entry(key)` (with preserve_order)
- `.shift_remove_entry(key)` (with preserve_order)

But poor `serde_json::map::OccupiedEntry` only has `.remove()`.

This means we can't easily/efficiently get the owned key from an `OccupiedEntry` - we have to either borrow and clone it, or call `.remove_entry(key)` on the original map, which will re-hash the key. Neither of these are catastrophic, but it's extra work we can avoid, since the underlying `OccupiedEntry` for both `BTreeMap` and `IndexMap` have a `remove_entry` method.

Similarly, with preserve_order enabled, we can't easily remove the entry without messing up the order of later entries, since `OccupiedEntry::remove()` uses `swap_remove` rather than `shift_remove`.

Hence, this PR adds the five missing removal methods to `serde_json::map::OccupiedEntry`

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -900,6 +900,69 @@ impl<'a> OccupiedEntry<'a> {
         #[cfg(not(feature = "preserve_order"))]
         return self.occupied.remove();
     }
+
+    /// Removes the entry from the map, returning the stored key and value.
+    ///
+    /// If serde_json's "preserve_order" is enabled, `.remove_entry()` is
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thanks!

PR #1182, opened 2024-08-23
Ensure the SWAR chunks are 64-bit in more cases

CryZe:
Various architectures have native support for 64-bit integers, but there are Rust targets for those architectures where the pointer size is intentionally just 32-bit. For SWAR this smaller pointer size would negatively affect those targets, so this PR ensures the chunk size stays 64-bit on those targets.

The same is done in various other crates, such as [hashbrown](https://github.com/rust-lang/hashbrown/blob/f1ba3b4547f5a8f0d4b679e4f44298138d89be59/src/raw/generic.rs#L10-L15). Though I wish there was some sort of `#[cfg(target_integer_width = "64")]`.


purplesyringa:
For context:

- `aarch64` is for target `aarch64-unknown-linux-gnu_ilp32`
- `x86_64` is for target `x86_64-unknown-linux-gnux32`

WASM's self-explanatory.

dtolnay (approved):
Thanks!

PR #1188, opened 2024-09-04
Add support for 128 bit HashMap key serialization

Mrreadiness:
Now `HashMap` doesn't support i128/u128 key serialization (deserialization works). 
Example of code:
```rust
let mut b = HashMap::new();
b.insert(0_i128, String::new());
serde_json::to_value(b).unwrap();
```
Error:
```bash
Error("i128 is not supported", line: 0, column: 0)
```
I don't find any reason why. So I added the missing methods.


dtolnay (approved):
Thank you!

PR #1199, opened 2024-10-17
Add methods on Map and Value to destroy preserved ordering

### LINKED ISSUE #1014: Add a `.sort()` method to expose `IndexMap::sort_keys()`
There isn't a good way that I know of to get sorted output when `preserve_order` is enabled (and disabling it isn't an option), which is important for getting canonical JSON and for comparing JSON strings. `IndexMap` has a `sort_keys`, could this be exposed somehow?

I'm imagining `Value::sort(&mut self)` and `Map::sort(&mut self)`, either of which would just walk the tree and call the methods on all nested items. When `preserve_order` is disabled, this would be a nop.

dtolnay:
Closes #1014.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -348,6 +348,28 @@ impl Map<String, Value> {
     {
         self.map.retain(f);
     }
+
+    /// Sorts this map's entries in-place using `str`'s usual ordering.
+    ///
+    /// If serde_json's "preserve_order" feature is not enabled, this method
+    /// does no work because all JSON maps are always kept in a sorted state.
+    ///
+    /// If serde_json's "preserve_order" feature is enabled, this method
+    /// destroys the original source order or insertion order of this map in
+    /// favor of an alphanumerical order that matches how a BTreeMap with the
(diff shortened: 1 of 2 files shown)


src/map.rs
```diff
@@ -348,6 +348,28 @@ impl Map<String, Value> {
...
+    ///
+    /// If serde_json's "preserve_order" feature is enabled, this method
+    /// destroys the original source order or insertion order of this map in
+    /// favor of an alphanumerical order that matches how a BTreeMap with the
+    /// same contents would be ordered. This takes **O(n log n + c)** time where
+    /// _n_ is the length of the map and _c_ is the capacity.
+    ///
+    /// Other maps nested within the values of this map are not sorted. If you
+    /// need the entire data structure to be sorted at all levels, you must also
+    /// call `map.`[`values_mut`]`().for_each(`[`Value::sort_all_objects`]`)`.
```
dtolnay, line 365: This is rendered badly but I filed https://github.com/rust-lang/rust/issues/131845 to follow up in rustdoc.

PR #1211, opened 2024-11-08
Forward `serde_json::value::RawValue` to `serde_json::RawValue`

bheylin:
Given that `Value` is in the root of the crate I argue that `RawValue` should be at the same level.

I use `RawValue` even more than I use `Value` and I would like to see it promoted, even if that just means
making the import more inviting to use.


dtolnay (commented):
Thanks for the PR. I would prefer to be pretty selective about what goes in the crate root. Needing RawValue more often than Value is not typical.

PR #1213, opened 2024-11-14
Add a safety comment to unsafe block

djmitche:
I noted the lack of a comment when performing a review. The implementation is safe, just not documented.


dtolnay (approved):
Thanks!

PR #1217, opened 2024-11-20
Improve error messages when deserializing enums

avandecreme:
This is a fix for https://github.com/serde-rs/serde/issues/2702

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -1881,7 +1881,14 @@ impl<'de, R: Read<'de>> de::Deserializer<'de> for &mut Deserializer<R> {
             Some(b'{') => {
                 check_recursion! {
                     self.eat_char();
-                    let ret = visitor.visit_enum(VariantAccess::new(self));
+                    let ret = match tri!(self.parse_whitespace()) {
+                        Some(b'}') => Err(self.fix_position(de::Error::invalid_value(
+                            Unexpected::Other("empty map"),
+                            &"enum variant",
+                        ))),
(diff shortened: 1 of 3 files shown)


vegardgs-ksat:
This PR also closes #1224

avandecreme:
@dtolnay Any interest in this MR?

src/de.rs
```diff
@@ -1882,12 +1889,11 @@ impl<'de, R: Read<'de>> de::Deserializer<'de> for &mut Deserializer<R> {
                         self.eat_char();
                         Ok(value)
                     }
-                    Some(_) => Err(self.error(ErrorCode::ExpectedSomeValue)),
+                    Some(_) => Err(self.error(ErrorCode::ExpectedObjectCommaOrEnd)),
```
robsdedude, line 1892: Nit pick: the error code isn't 100% correct. It states that `,` was expected `}`. However, only `}` is valid here.

Here's an example of that
```rust
#[derive(Deserialize, Debug)]
enum Foo {
    Bar(u64, bool),
}

let data = r#"{"Bar": [1234, true],}"#;
let res = from_str::<Foo>(data);
dbg!(res);
```
Yields
```
res = Err(
    Error("expected `,` or `}`", line: 1, column: 20),
)
```
So the error is complaining that the `,` isn't a `,` or `}` :upside_down_face:

avandecreme: Good point. Fixed

PR #1221, opened 2024-12-11
Add literal 'null', 'true' and 'false' consts to `RawValue` struct.

bheylin:
I have been using `RawValue` a lot lately to implement server endpoints that accept large JSON objects, extract some specific fields and  ultimately save the entire JSON to the DB.

As part of this process, a proc-macro based JSON pre-processor is used to automate HTTP error responses. This also uses `RawValue` to avoid allocating when traversing unknown JSON data.

The pre-processor uses literal "null"'s as part of it's logic, and at the moment I'm deserializing a literal `"null"` string to `RawValue` every time I need one.

This could be done once and stored in a `LazyLock` or something similar, but I thought that adding literal values to the `RawValue` struct is the least complex solution.

What do you think?

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -120,11 +120,11 @@ pub struct RawValue {
 
 impl RawValue {
     /// A literal JSON null value as `RawValue`.
-    pub const NULL: &RawValue = RawValue::from_borrowed("null");
+    pub const NULL: &'static RawValue = RawValue::from_borrowed("null");
     /// A literal JSON boolean true value as `RawValue`.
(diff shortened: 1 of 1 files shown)


dtolnay:
Published in serde_json 1.0.134.

I was waiting to publish this because I was not happy with how rustdoc originally rendered these. https://github.com/rust-lang/rust/pull/134321

bheylin:
No worries, there's no rush on a release for me.

I didn't realize that Rustdoc rendered each impl literally.

PR #1228, opened 2025-01-17
Add generic key-value support for Map

HoKim98:
Hello, this PR supports `Map<K, V>`.

Generic maps are useful for deserializing maps (i.e. `impl<'de> Deserializer<'de, K, V> for Map<K, V>`) into other types.

## Side-effects

The map `Map<K, V>` implements other map-like operations and traits.
It may ask for users to specify the types `K, V`.

```rust
let mut map = serde_json::Map::new();  // error[E0282]: type annotations needed for `serde_json::Map<_, _>`
assert_eq!(map.entry("serde").key(), &"serde");
```

Instead, I set the default type of `Map` to `Map<String, Value>`.

```rust
let mut map: serde_json::Map = serde_json::Map::new();
assert_eq!(map.entry("serde").key(), &"serde");
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -15,6 +15,7 @@ use core::borrow::Borrow;
 use core::fmt::{self, Debug};
 use core::hash::{Hash, Hasher};
 use core::iter::FusedIterator;
+use core::marker::PhantomData;
 #[cfg(feature = "preserve_order")]
 use core::mem;
 use core::ops;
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks for the PR. I would prefer to keep the less generic implementation so that users are not forced to write type annotations so often when using serde_json::Map. This type is only intended for representing a map deserialized from JSON. For any other map, please use a different map type.

HoKim98:
Then what about renaming `Map` to other one (e.g. `GenericMap`) and make an alias for `Map` (i.e. `type Map = GenericMap<String, Value>;`)? It minimizes the side effect because there is few usage of `Map<K, V>` for now.

PR #1230, opened 2025-01-18
Use Map::with_capacity in value::Serializer::serialize_map

goffrie:
Much like how serialize_seq uses Vec::with_capacity.


dtolnay (approved):
Thanks!

PR #1250, opened 2025-03-20
Add feature "sort_arrays" and API for sotring all arrays

azoyan:
Add special feature `sort_arrays` and function to sort object and arrays.

It is necessary for storing JSON and comparing with previous JSON. For example for prevent DDOS attack where attacker reorder elements in array and create new request body and avoid block by rate limit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "serde_json"
 version = "1.0.140"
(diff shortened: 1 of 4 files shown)


dtolnay (commented):
I would prefer not to support this. But this functionality can be implemented in any downstream project that needs it.

PR #1256, opened 2025-04-26
Feature valuable

### LINKED ISSUE #1255: Implement valuable::Valuable on serde_json::Value as a feature flag
I am wondering if we can implement this trait from the valuable crate on serde_json::Value. 

This would be really helpful with tracing and logging the serde_json::Value as a structured json. 

I found this [pr](https://github.com/tokio-rs/valuable/pull/61) from 4 years ago in the valuable crate but never been merged and as the description says it should be included in this crate.

If this is something that could be merged, I wouldn't mind working on this.

Thanks!

Vrajs16:
Adds support for valuable crate as feature flag

closes #1255

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@miri
+        with:
+          toolchain: nightly-2025-05-16  # https://github.com/rust-lang/miri/issues/4323
       - run: cargo miri setup
       - run: cargo miri test --target ${{matrix.target}}
       - run: cargo miri test --target ${{matrix.target}} --features preserve_order,float_roundtrip,arbitrary_precision,raw_value
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,7 @@
 [package]
 name = "serde_json"
-version = "1.0.140"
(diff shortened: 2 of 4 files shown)


gnosek:
> Not sure how to pass the rust 1.56.1 check, if someone could help with that

One way to do this is to use a different feature name so that it doesn't conflict with the crate name.

rollo-b2c2:
Any progress on this?

Would be really useful, right now I'm just copying this wholesale into my source tree (with new types)

michaeltlombardi:
If there's any remaining work or concerns for this PR, I would be happy to tackle it. A project I'm currently working on has a need for consuming arbitrary trace data from external programs and we're already using JSON Lines for emitting messages that our program can bubble up. Unfortunately, because we can't implement `Valuable` on `serde_json::Value` ourselves, we would need to use the newtype pattern to wrap the incoming data (and also unwrap it if we add support for more known structured data fields so we can validate them correctly).

I previously offered to adopt the implementation in the `valuable` crate (see PR: tokio-rs/valuable#61) but haven't heard back from the maintainers. I don't have very strong opinions about whether this should be implemented in `tracing` or this crate, but I do think that there's strong value in _one_ of the libraries closing this work out to enabling structured tracing given the ubiquity of JSON both as a common interchange format and as a specific export format for trace messages.

PR #1264, opened 2025-05-31
Add Clone and Debug impls to map iterators

xlambein:
I've noticed that the iterators for `Map` are missing some traits that are implemented by their counterpart in `std::collections` and `indexmap`: namely, `Clone` and `Debug`. AFAICT, there's no reason for those to be missing, and they can be useful. I found myself missing `Clone` on `Iter`, for example.

Initially, I'd included `Default` among these, but it's not supported on Rust 1.65 and older.


dtolnay (approved):
Thanks!

PR #1265, opened 2025-06-18
feat: impl Default for &Value

aatifsyed:
A rounded edge for:
```
let val: Value;
val.pointer("/foo).unwrap_or_default()
```

Obviously not required.


aatifsyed:
@dtolnay feel free to close this as WONTFIX :)

dtolnay (approved):
Thank you!

PR #1269, opened 2025-07-08
Add `Object` type alias for JSON objects

### LINKED ISSUE #776: Provide an Object type alias
I find myself writing

```rust
type JsonObject = serde_json::Map<String, serde_json::Value>;
```

again and again. Would you accept a PR that adds a type alias like that (without the `Json` prefix) to `serde_json::value`, uses it for the `Value::Object` variant and maybe re-exports it at the crate root too?

CryZe:
This PR adds a convenient `Object` type alias to eliminate the need for repeatedly writing `Map<String, Value>` when working with JSON objects.

Instead of:
```rust
use serde_json::{Map, Value};
let obj: Map<String, Value> = serde_json::from_str(json_str)?;
```

Users can now write:
```rust
use serde_json::Object;
let obj: Object = serde_json::from_str(json_str)?;
```

Fixes #776

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -402,6 +402,9 @@ pub use crate::ser::{to_writer, to_writer_pretty, Serializer};
 #[doc(inline)]
 pub use crate::value::{from_value, to_value, Map, Number, Value};
 
+/// Represents a JSON object.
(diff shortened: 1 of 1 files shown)


jplatte:
Ping @dtolnay

PR #1273, opened 2025-07-18
string serialization escaping optimisations

conradludgate:
While serializing strings, we enter the hot loop checking 1 byte at a time. By massaging the code and using a little bit of unsafe, we can optimise this hot loop a bit.

<details>
<summary>This benchmark shows about a 2-5% speedup when working with string heavy documents</summary>

Benchmark was run with target x86_64-unknown-linux-gnu on an Intel Xeon Platinum 8375C CPU @ 2.90GHz, as well as on an Apple M4 Max.

```rust
use std::hint::black_box;

use criterion::{criterion_group, criterion_main, Criterion};
use serde_json::Value;

pub fn k8s(c: &mut Criterion) {
    // https://raw.githubusercontent.com/kubernetes/kubernetes/v1.33.3/api/openapi-spec/swagger.json
    let k8s = std::fs::read_to_string("benches/k8s-openapi.json").unwrap();
    let value: Value = serde_json::from_str(&k8s).unwrap();
    drop(k8s);

    let mut v = Vec::new();
    serde_json::to_writer_pretty(&mut v, &value).unwrap();

    c.bench_function("pretty", |b| {
        b.iter(|| {
            v.clear();
            serde_json::to_writer_pretty(&mut v, &value).unwrap();
            black_box(&v[..]);
        })
    });

    c.bench_function("compact", |b| {
        b.iter(|| {
            v.clear();
            serde_json::to_writer(&mut v, &value).unwrap();
            black_box(&v[..]);
        })
    });
}

criterion_group!(benches, k8s);
criterion_main!(benches);
```
</details>

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -1534,23 +1534,6 @@ pub enum CharEscape {
     AsciiControl(u8),
 }
 
-impl CharEscape {
-    #[inline]
-    fn from_escape_table(escape: u8, byte: u8) -> CharEscape {
-        match escape {
-            self::BB => CharEscape::Backspace,
-            self::TT => CharEscape::Tab,
-            self::NN => CharEscape::LineFeed,
-            self::FF => CharEscape::FormFeed,
-            self::RR => CharEscape::CarriageReturn,
-            self::QU => CharEscape::Quote,
-            self::BS => CharEscape::ReverseSolidus,
-            self::UU => CharEscape::AsciiControl(byte),
-            _ => unreachable!(),
-        }
-    }
-}
-
 /// This trait abstracts away serializing the JSON control characters, which allows the user to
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks!

Would there be any benefit to coordinating the discriminants of `CharEscape` with the values that are being converted to/from it in this code, even if `as` cannot be used?

```rust
#[repr(u8)]
pub enum CharEscape {
    Quote = b'"',
    ReverseSolidus = b'\\',
    Solidus = b'/',
    Backspace = b'b',
    FormFeed = b'f',
    LineFeed = b'n',
    CarriageReturn = b'r',
    Tab = b't',
    AsciiControl(u8) = b'u',
}
```

dtolnay (approved):

PR #1291, opened 2025-10-15
Fix missing escape character lookup for forward slash (solidus).

patrickcjh:
Forward slash is missing in the `ESCAPE` lookup table.


dtolnay (commented):
I think this is fine as is. `/` is not required to be escaped in JSON strings.

PR #1295, opened 2025-11-04
chore: minor improvement for docs

claudecodering:
minor improvement for docs


dtolnay (commented):
Thanks. I am merging the 'a'->'an' change without the comma changes, as I think those sentences are fine as is.

PR #1324, opened 2026-05-21
Reject non-string enum object keys

### LINKED ISSUE #979: `EnumAccess.variant` can parse non-string JSON object member names
### Version
1.0.91

### Description
When `visit_enum` is used for a JSON object, then `EnumAccess.variant` can parse non-string JSON object member names, which is malformed JSON. The reason for this is likely that the following line does not enforce that the next value is a string:
https://github.com/serde-rs/json/blob/e41ee42d92022dbffc00f4ed50580fa5e060a379/src/de.rs#L2019

### Reproducer code
```rust
use serde::de::{Deserializer, VariantAccess, Visitor};

struct V;
impl<'de> Visitor<'de> for V {
    type Value = (Vec<bool>, ());

    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(formatter, "enum")
    }

    fn visit_enum<A>(self, data: A) -> Result<Self::Value, A::Error>
    where
        A: serde::de::EnumAccess<'de>,
    {
        // Issue is here: Implementation for serde_json's internal VariantAccess struct
        // reads object member as regular value, without enforcing that it is a string
        let (key, variant_access) = data.variant()?;
        let value = variant_access.newtype_variant()?;
        Ok((key, value))
    }
}

fn main() {
    let mut de = serde_json::Deserializer::from_str("{[true]: null}");
    let entry = de.deserialize_enum("name", &[], V).unwrap();
    println!("entry: {entry:?}");
    de.end().unwrap();
}

```
This deserializes the malformed JSON object `{[true]: null}` without reporting any error.

puneetdixit200:
## Summary
- validate that object-form enum variants start with a JSON string key
- deserialize enum object keys through the same `MapKey` path used by map keys
- add a regression test for malformed `{[true]: null}` input

Fixes #979

## Tests
- `cargo fmt --check`
- `cargo test --test regression issue979`
- `cargo test --test test test_parse_enum_errors`
- `cargo test`
- `cargo clippy --all-targets -- -D warnings`

### DIFF
--- a/src/de.rs
+++ b/src/de.rs
@@ -2052,7 +2052,17 @@ impl<'de, 'a, R: Read<'de> + 'a> de::EnumAccess<'de> for VariantAccess<'a, R> {
     where
         V: de::DeserializeSeed<'de>,
     {
-        let val = tri!(seed.deserialize(&mut *self.de));
+        match tri!(self.de.parse_whitespace()) {
+            Some(b'"') => {}
+            Some(b'}') => return Err(self.de.peek_error(ErrorCode::ExpectedSomeValue)),
+            Some(_) => return Err(self.de.peek_error(ErrorCode::KeyMustBeAString)),
+            None => return Err(self.de.peek_error(ErrorCode::EofWhileParsingObject)),
+        }
+
+        let val = match seed.deserialize(MapKey { de: &mut *self.de }) {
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1329, opened 2026-07-01
Handle collect_str Display errors without panicking

### LINKED ISSUE #1328: `serde_json::Serializer::collect_str` can panic if `Display` returns `fmt::Error` directly
`serde_json`'s serializer has a panic path in `collect_str`: if the value's `Display` implementation returns `fmt::Error` without the serializer adapter recording an underlying I/O error, serde_json unwraps `adapter.error` with `expect`.

This is a lower-severity robustness issue because a `Display` implementation returning `fmt::Error` directly is unusual. Still, the reproducer uses safe public APIs and the panic is not documented near the serialization path.

Version checked: `serde_json 1.0.150`

## Relevant code

```rust
let mut adapter = Adapter {
    writer: &mut self.writer,
    formatter: &mut self.formatter,
    error: None,
};
match write!(adapter, "{}", value) {
    Ok(()) => debug_assert!(adapter.error.is_none()),
    Err(fmt::Error) => {
        return Err(Error::io(adapter.error.expect("there should be an error")));
    }
}
```

The adapter records `adapter.error` when serde_json's own writer/formatter returns an I/O error. But if the caller-provided `Display` implementation returns `fmt::Error` directly, `adapter.error` is still `None`.

## Reproducer

```rust
use std::fmt;
use serde::ser::{Serialize, Serializer};

struct BadDisplay;

impl fmt::Display for BadDisplay {
    fn fmt(&self, _f: &mut fmt::Formatter<'_>) -> fmt::Result {
        Err(fmt::Error)
    }
}

impl Serialize for BadDisplay {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        serializer.collect_str(self)
    }
}

#[test]
#[should_panic(expected = "there should be an error")]
fn collect_str_display_error_panics() {
    let _ = serde_json::to_string(&BadDisplay);
}
```

Observed behavior:

```text
thread '...' panicked at src/ser.rs:451:52:
there should be an error
```

## Expected behavior

`to_string` should return a serialization error instead of panicking, or the panic condition should be documented if serde_json intentionally assumes `Display` never returns `fmt::Error` directly.

<details>
<summary>Lower-priority formatter state-machine cases</summary>

The generated tests also included direct calls to `PrettyFormatter::end_array` and `PrettyFormatter::end_object` without a matching begin call:

```rust
use serde_json::ser::{Formatter, PrettyFormatter};

#[test]
#[should_panic]
fn end_array_without_begin_panics() {
    let mut formatter = PrettyFormatter::new();
    let mut out = Vec::new();
    Formatter::end_array(&mut formatter, &mut out).unwrap();
}
```

These underflow `current_indent` in debug builds:

```rust
fn end_array<W>(&mut self, writer: &mut W) -> io::Result<()>
where
    W: ?Sized + io::Write,
{
    self.current_indent -= 1;
    /* ... */
}
```

I would not lead with these because they violate the formatter state-machine ordering that serde_json's serializer normally maintains. They are worth mentioning only as optional hardening/documentation notes.

</details>

sjh9714:
## Summary

Fixes #1328.

`Serializer::collect_str` currently assumes every `fmt::Error` came from serde_json's writer adapter, so a user `Display` implementation that returns `fmt::Error` panics when no writer error was captured. This returns a serde data error for that case while preserving the existing IO-error path when the adapter records one.

## Changes

- Convert user-originated `fmt::Error` from `collect_str` into a serde serialization error instead of unwrapping a missing adapter error.
- Add a regression test for `Display` returning `fmt::Error` through `collect_str`.

## Testing

- `PATH="$HOME/.cargo/bin:$PATH" cargo test --test regression issue1328` (fails before this fix with `there should be an error`)
- `PATH="$HOME/.cargo/bin:$PATH" cargo test --test regression issue1328`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --check`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test --test regression`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test`
- `git diff --check`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/src/ser.rs
+++ b/src/ser.rs
@@ -448,7 +448,14 @@ where
         match write!(adapter, "{}", value) {
             Ok(()) => debug_assert!(adapter.error.is_none()),
             Err(fmt::Error) => {
-                return Err(Error::io(adapter.error.expect("there should be an error")));
+                return Err(adapter.error.map_or_else(
+                    || {
+                        <Error as ser::Error>::custom(
+                            "Display implementation returned an error unexpectedly",
+                        )
+                    },
+                    Error::io,
+                ));
             }
         }
         self.formatter
--- a/tests/regression/issue1328.rs
+++ b/tests/regression/issue1328.rs
@@ -1,0 +1,30 @@
+use serde::ser::{Serialize, Serializer};
+use std::fmt;
+
+struct BadDisplay;
+
+impl fmt::Display for BadDisplay {
+    fn fmt(&self, _: &mut fmt::Formatter<'_>) -> fmt::Result {
+        Err(fmt::Error)
+    }
+}
+
+impl Serialize for BadDisplay {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.collect_str(self)
+    }
+}
+
+#[test]
+fn test() {
+    let err = serde_json::to_string(&BadDisplay).unwrap_err();
+
+    assert!(err.is_data());
+    assert_eq!(
+        err.to_string(),
+        "Display implementation returned an error unexpectedly"
+    );
+}


dtolnay (commented):
I believe the original code is correct as written. See https://github.com/serde-rs/json/issues/1328#issuecomment-4868449994.

Thanks anyway!

PR #1334, opened 2026-07-28
Remove deprecated package authors field from Cargo.toml

xtqqczze:
`package.authors` is deprecated: https://github.com/rust-lang/cargo/pull/15068


dtolnay (approved):
Thanks!

